PR #3, opened 2022-04-18
Enhancement/remove lodash dependency

AGalabov:


src/lib/lodash.ts
```diff
@@ -0,0 +1,55 @@
+export function isUndefined<T>(value: any): value is undefined {
```
georgyangelov, line 1: Not sure if we need a function for that 🤷 We can just write `_ => _ === undefined` wherever

src/lib/lodash.ts
```diff
@@ -0,0 +1,55 @@
...
+  return value === undefined;
+}
+
+export function isNil<T>(value: any): value is null | undefined {
+  return value == null;
```
georgyangelov, line 6: I'd write this out explicitly instead of using double-equals

src/lib/lodash.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export function mapValues<
+  T extends object,
+  MapperResult,
+  Result = { [K in keyof T]: MapperResult }
+>(object: T, mapper: (val: T[keyof T]) => MapperResult): Result {
+  const result = {} as Result;
+
+  Object.entries(object).forEach(([key, value]: [string, T[keyof T]]) => {
+    // @ts-ignore
```
georgyangelov, line 17: еееееее ееееееее

src/lib/lodash.ts
```diff
@@ -0,0 +1,55 @@
...
+}
+
+export function mapValues<
+  T extends object,
+  MapperResult,
+  Result = { [K in keyof T]: MapperResult }
+>(object: T, mapper: (val: T[keyof T]) => MapperResult): Result {
+  const result = {} as Result;
+
+  Object.entries(object).forEach(([key, value]: [string, T[keyof T]]) => {
```
georgyangelov, line 16:
```
const result: any = {};

for (const [key, value] = Object.entries(object)) {
  result[key] = mapper(value);
}

return result;
```
?

src/lib/lodash.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export function omit<
+  T extends object,
+  Keys extends keyof T,
+  Result = Omit<T, Keys>
+>(object: T, keys: Keys[]) {
+  const result = {} as Result;
+
+  Object.entries(object).forEach(([key, value]) => {
+    if (keys.every((keyToOmit) => keyToOmit !== key)) {
```
georgyangelov, line 32: Would be more intuitive with `some`. Also can you get rid of the `@ts-ignore`s please, you can use an `any` if need be

src/lib/lodash.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export function omitBy<
+  T extends object,
+  Result = Partial<{ [K in keyof T]: T[keyof T] }>
+>(object: T, predicate: (val: T[keyof T]) => boolean): Result {
+  const result = {} as Result;
+
+  Object.entries(object).forEach(([key, value]) => {
+    if (!predicate(value)) {
+      // @ts-ignore
```
georgyangelov, line 49: :(

src/lib/lodash.ts
AGalabov: We can but if it's inside a filter for example it would not infer the type correctly iwthout the `is undefined` return type. Not sure if we need that in our use case but didn't hurt

src/lib/lodash.ts
AGalabov: I got tired, ngl. Sorry :shame:

src/lib/lodash.ts
AGalabov: `any` is the way

PR #4, opened 2022-04-19
added support for all type of responses

AGalabov:


src/openapi-generator.ts
```diff
@@ -86,6 +95,23 @@ export class OpenAPIGenerator {
     };
   }
 
+  private sortDefinitions() {
```
georgyangelov, line 98: I see now. This should be ok, my other idea was to do a pass like this:

```ts
this._definitions.forEach((def) => {
  switch (def.type) {
    case 'schema': this.schemaRefs[def.refId] = def; break;
    case 'parameter': this.paramRefs[def.refId] = def; break;
    case 'route': this.routeRefs[def.refId] = def; break;
  }
});
```

georgyangelov (approved):

src/openapi-generator.ts
```diff
@@ -382,6 +402,42 @@ export class OpenAPIGenerator {
...
+      return {
+        description: metadata.description,
+      };
+    }
+
+    const metadata = this.getMetadata(response.schema);
+    const responseSchema = this.generateInnerSchema(response.schema);
+
+    if (!metadata?.description) {
+      throw new Error(
```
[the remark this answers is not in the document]
AGalabov: Yes it is. Responses need the description key. Typescript would also complain - `ResponseObject` requires it correctly as well

src/openapi-generator.ts
```diff
@@ -151,42 +170,52 @@ export class OpenAPIGenerator {
       const parameters = Object.entries(propTypes).map(([key, schema]) => {
         const innerMetadata = this.getMetadata(schema);
 
-        if (innerMetadata?.name && innerMetadata.name !== key) {
+        const innerParameterMetadata = innerMetadata?.param;
```
georgyangelov, line 173: We should document this somewhere in the README

src/openapi-registry.ts
```diff
@@ -18,8 +18,9 @@ export interface RouteConfig extends Partial<OperationObject> {
     query?: ZodObject<any>;
     headers?: ZodType<unknown>[];
   };
-  response: ZodType<unknown>;
-  errors?: any[];
+  responses: {
+    [key: number]: ResponseConfig;
```
georgyangelov, line 22: Wasn't there a `default` kеy somewhere? Also JS object keys can only be strings, so this is a bit weird

src/openapi-generator.ts
AGalabov: Yup, I'll add it in the Readme.md. I'll probably separate them into:
1. Defining schemas
2. Defining parameters
3. Defining routes

src/openapi-registry.ts
AGalabov: Hmm there is a default key, yes. I guess it should be closer to the type `ResponsesObject`
```
    default?: ResponseConfig;
    [statuscode: string]: ResponseConfig;
```

AGalabov: And that's basically: `[status: string]: ResponseConfig;`

Issue #7, opened 2022-04-29
Parameters inlined instead of using `$ref`?


georgyangelov:
In [this example](https://github.com/asteasolutions/zod-to-openapi/blob/a554dac86bfd67e2072ddbaad8496a162dd10d5e/example/openapi-docs.yml#L42) the parameter is not using `$ref`. Why? Is this a docs issue or a library issue?

AGalabov:
Fixed in https://github.com/asteasolutions/zod-to-openapi/pull/8

PR #8, opened 2022-04-29
Bugfix #7 object parameters not referenced

AGalabov:


src/errors.ts
```diff
@@ -0,0 +1,58 @@
+export class BaseError {
```
georgyangelov, line 1: Can you call this `ZodToOpenAPIError`?

src/errors.ts
```diff
@@ -0,0 +1,58 @@
...
+  constructor(private message: string, private data?: unknown) {}
```
georgyangelov, line 2: I would not put an unknown `data` property, just add new props in subclasses. Since these errors will not be serialized anyway it's cleaner.

georgyangelov (approved):
It would have been fine to leave them as bare `Error` instances as well, since these are programmer errors, but it's fine like this as well, appreciate the effort :)

src/errors.ts
AGalabov: Not sure, man. I'll try my best :smile:

PR #30, opened 2022-09-03
Task/#29 add support for extended schemas

AGalabov:


spec/polymorphism.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+        ],
+      },
+    });
+  });
+
+  it.todo('can apply nullable');
+
+  it.todo('can apply optional');
+
+  it.todo('can apply openapi');
```
georgyangelov, line 56: I assume we'll do this if I like the implementation? :D

src/openapi-generator.ts
```diff
@@ -655,19 +655,52 @@ export class OpenAPIGenerator {
...
 
+    const properties = mapValues(propTypes, propSchema =>
+      this.generateInnerSchema(propSchema)
+    );
+
+    if (extendedFrom) {
+      const alreadyRegistered = Object.keys(
+        this.schemaRefs[extendedFrom]?.properties ?? {}
+      );
+      const alreadyRequired = this.schemaRefs[extendedFrom]?.required ?? [];
```
georgyangelov, line 675: If the schema does not exist in the registry maybe we should fall back to the "else" case instead of defaulting to empty arrays and letting it flow through the "extendedFrom" logic?

src/openapi-generator.ts
```diff
@@ -655,19 +655,52 @@ export class OpenAPIGenerator {
...
+      );
+
+      const additionalData = {
+        type: 'object' as const,
+
+        properties: additionalProperties,
+
+        required:
+          additionallyRequired.length > 0 ? additionallyRequired : undefined,
+      };
```
georgyangelov, line 689: We're skipping the `additionalProperties` check here. Can we somehow merge the two branches actually? e.g. if (extendedFrom), then just set alreadyRegistered and alreadyRequired and continue with the same logic regardless, just default those to empty, then wrap in `allOf` at the end if necessary?

src/zod-extensions.ts
```diff
@@ -42,6 +44,23 @@ export function extendZodWithOpenApi(zod: typeof z) {
         },
       },
     });
+
+    if (isZodType(this, 'ZodObject')) {
+      const initialExtend = this.extend;
+
+      // TODO: This does an overload everytime. So .extend().openapi() makes this change twice
+      result.extend = function (...args: any) {
```
georgyangelov, line 52: Actually wouldn't this be applied to an object only if it already had `openapi` called? Is that the intention? Why not override it globally and only use it if necessary?

src/zod-extensions.ts
```diff
@@ -42,6 +44,23 @@ export function extendZodWithOpenApi(zod: typeof z) {
...
+        const extendedResult = initialExtend.apply(result, args);
+
+        extendedResult._def.openapi = {
+          extendedFrom: result._def.openapi?.refId,
+        };
```
georgyangelov, line 57: Aren't we losing the other keys here?

spec/polymorphism.spec.ts
AGalabov: Yes, that is the plan. However the `nullable` and `optional` ones might be done in a separate PR since they are related to another problem (passing `nullable` on an already generated and referenced schema) - see the `it.skip` bellow

src/openapi-generator.ts
AGalabov: I see your point. This code shouldn't actually ever occur to be honest unless someone passes `extendedFrom` manually or the order is messed up somehow. I think I am more inclined to throw an error instead :thinking: .

src/openapi-generator.ts
AGalabov: You are correct about `additionalProperties` I think. But for `nullable` it is more complicated from what I noticed:
```yaml
allOf:
  - $ref: <some-ref>
  - type: object
     properties:
        a:
          type: string
   nullable: true # This right here is wrong/does not produce any result
```
The above is based on the [docs](https://swagger.io/docs/specification/using-ref/)
image
I believe that if you are adding a nullable the result should be something like:
```yaml
anyOf:
  - type: 'null' # I am still unable to confirm that this is a valid syntax
  - allOf: 
     - $ref: <some-ref>
     - type: object
       properties:
          a:
            type: string
```

So basically I don't think we can join them as they need to be 2 separate flows

src/zod-extensions.ts
AGalabov: Overriding it globabally was the initial idea, yes. The problem there was that if we look at the zod implementation we have the following:
```js
class ZodObject extends ZodType {
  constructor() {
      super(...arguments);
      // some other stuff
      this.extend = AugmentFactory(this._def);
   }
```
This means that we cannot actually modify `zod.ZodObject.prototype.extend` since it is not a "static" (as in already exists on the class) method but is being generated upon object construction.

That being said I still think this approach is elegant since in order to "create a refId" (be it manually or from the registry) you need to call the `.openapi` method. So in some sense we would only modify objects that are being used with our library.

src/zod-extensions.ts
AGalabov: I was thinking about for some time. What I got to was something along the lines of:
1. We have an object with its metadata added into a refference
2. We create a new object with extend. This is the creation of a completely separate instance that is not meant to have the parent object's metadata since it would use `allOf` in the docs.
3. This is like a constructor so it should not be possible to have any other relevant metadata inside of it. The `refId` that it would get registered by comes after the schema is generated so they do not interfere with each other.

And yet at the same time I did have my inner doubts. I'll see if there is any scenario that I can think of, where this won't work. Feel free to help me with this, tho :smile:

georgyangelov (approved):

PR #33, opened 2022-09-23
Expose the RouteConfig interface (proposal)

### LINKED ISSUE #32: Expose RouteConfig from openapi-registry
I am looking to build helper functions for controlling access to the registry, and having all the logic centralized.

I'd like my helper function to take in the RouteConfig as a parameter. Having access to your RouteConfig would be helpful here, so I don't have to mimic the type.

I made a quick edit in my local code base and it seems to work fine, by just adding RouteConfig to the export.

Let me know what you think! Thanks!

https://github.com/asteasolutions/zod-to-openapi/blob/fdf240a40182d2c8cd7797bc891f67d3c2a04314/src/index.ts#L3

Fish1:
PR Example of what I am looking for.

closes #32

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,3 +1,3 @@
 export * from './zod-extensions';
 export { OpenAPIGenerator } from './openapi-generator';
-export { OpenAPIRegistry, RouteConfig } from './openapi-registry';
+export { OpenAPIRegistry, RouteConfig, ResponseConfig } from './openapi-registry';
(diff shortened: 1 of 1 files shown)


AGalabov:
Looks good

Fish1:
I've only exposed the interfaces I personally needed at the moment. Do you think it would be helpful to expose all the parameter interfaces?

A common patter I use is...
```ts
const myParam: Type = {
  blah: 1,
  test: 2,
}

doStuff(myParam);
```

So maybe just exposing all of the parameters would be a nice to have.

AGalabov:
@Fish1 I am okay with exporting `ResponseConfig` as well. The others are more of an internal implementation types.

AGalabov:
@Fish1 I see that the build checks are failing. Can you check if the code is formatted properly. You can check try and run `npm run lint`. I imagine that the export would be reordered into a multiline one.

Fish1:
I checked the linter and ran prettier to properly format the code.

PR #36, opened 2022-09-29
#31 add support for multiple response media types

### LINKED ISSUE #31: Multiple schemas for response code
Does zod-to-openapi support different schema's for different content types but the same response code?

Such as
```
200:
  application/json:
    schema:
      $ref: "#/components/jsonResponse"
  text/csv:
    schema:
      $ref: "#/components/csvResponse"
```

### LINKED ISSUE #34: Request body only allows 'application/json'
Currently we only allow request bodies with media type `application/json`

I would like to use `application/x-www-form-urlencoded`. Unfortunately these are still quite common, including in the [OAuth spec](https://www.rfc-editor.org/rfc/rfc6749#section-4.4.2). 

`application/json` is hard-coded here:
https://github.com/asteasolutions/zod-to-openapi/blob/5fc40ede62ec4891a899469d941f20eb11ea26a2/src/openapi-generator.ts#L423-L425

I'm not sure how to modify the existing openapi extension to stay consistent with the current design.

AGalabov:
* Added support for multiple media types for request body and responses
* closes #31 
* closes #34

### DIFF
--- a/spec/setup-tests.ts
+++ b/spec/setup-tests.ts
@@ -1,0 +1,6 @@
+import { z } from 'zod';
+import { extendZodWithOpenApi } from '../src/zod-extensions';
+
+console.log('Executing file');
+
+extendZodWithOpenApi(z);
--- a/spec/setup-tests.ts
+++ b/spec/setup-tests.ts
@@ -1,6 +1,4 @@
 import { z } from 'zod';
 import { extendZodWithOpenApi } from '../src/zod-extensions';
 
-console.log('Executing file');
-
 extendZodWithOpenApi(z);
--- a/src/openapi-registry.ts
+++ b/src/openapi-registry.ts
@@ -1,7 +1,9 @@
 import {
   CallbackObject,
   ComponentsObject,
+  EncodingObject,
   ExampleObject,
+  ExamplesObject,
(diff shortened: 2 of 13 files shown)


spec/setup-tests.ts
```diff
@@ -0,0 +1,6 @@
+import { z } from 'zod';
+import { extendZodWithOpenApi } from '../src/zod-extensions';
+
+console.log('Executing file');
```
georgyangelov, line 4: Remove this?

src/openapi-registry.ts
```diff
@@ -14,26 +16,39 @@ import {
...
-  | {
-      mediaType: string;
-      schema: ZodType<unknown>;
-      description?: string;
-      headers?: HeadersObject;
-      links?: LinksObject;
-    }
-  | ZodVoid;
+export interface ZodMediaTypeObject {
+  schema: ZodType<unknown>;
```
georgyangelov, line 24: Do we want to allow custom schemas to be passed here instead of only Zod schemas? I'm not sure how response formats which are not json could be described using Zod schemas. For example, XML attributes?

georgyangelov (approved):

PR #37, opened 2022-10-05
Add support for integers

samchungy:
Hey legends, love this library so far. Just a small 🤏 PR to map numbers to integers when `.int()` is used.


georgyangelov (approved):
Thanks!

PR #38, opened 2022-10-06
Introduce string pattern and formats

samchungy:
I might need to address some test file conflicts with the last PR.

This PR adds some `format` mappings for some known types like `uuid`, `email` and `url` as well as takes any regex pattern from `.regex()` and adds it to the `pattern` field.

### DIFF
--- a/src/openapi-generator.ts
+++ b/src/openapi-generator.ts
@@ -505,10 +505,10 @@ export class OpenAPIGenerator {
   }
 
   private getZodStringCheck<T extends ZodStringDef['checks'][number]['kind']>(
-    zodSchema: ZodString,
+    zodString: ZodString,
     kind: T
   ) {
-    return zodSchema._def.checks.find(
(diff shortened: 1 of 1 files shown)


spec/simple.spec.ts
```diff
@@ -1,10 +1,40 @@
...
+        SimpleString: { type: 'string' },
+      });
+    });
+
+    it.each`
+      format     | zodString             | expected
+      ${'uuid'}  | ${z.string().uuid()}  | ${'uuid'}
+      ${'email'} | ${z.string().email()} | ${'email'}
+      ${'url'}   | ${z.string().url()}   | ${'uri'}
+    `(
```
georgyangelov, line 17: Hah, didn't know this existed

georgyangelov (approved):
Looks good, thank you!

AGalabov (commented):
Overall looks great. Thank you! Left 2 very small comments.

PR #41, opened 2022-10-07
Treat omit/pick as new objects

samchungy:
When I gave something a `refId` and I used `omit` and `pick` on that object. The resulting schema was still trying to use the existing `refId` which is incorrect. This treats new objects created with `.omit` and `.pick` as new objects essentially by clearing the `_def.openapi` from the object. Not sure if there's a better way to do this?

### DIFF
--- a/spec/polymorphism.spec.ts
+++ b/spec/polymorphism.spec.ts
@@ -135,7 +135,7 @@ describe('Polymorphism', () => {
     });
   });
 
-  it('correctly creates references to existing schema objects using refId', () => {
+  it('treats objects created by .omit as a new object', () => {
     const BaseSchema = z
       .object({
         name: z.string(),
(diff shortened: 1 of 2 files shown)


spec/polymorphism.spec.ts
```diff
@@ -134,4 +134,106 @@ describe('Polymorphism', () => {
       },
     });
   });
+
+  it('correctly creates references to existing schema objects using refId', () => {
```
AGalabov, line 138: We already have multiple tests for that. You can check:
* it('creates separate schemas and links them' ...
* it('supports records with refs' ...
* it('supports nullable for registered schemas' ...
So I don't think this test verifies anything new, does it?

spec/polymorphism.spec.ts
```diff
@@ -134,4 +134,106 @@ describe('Polymorphism', () => {
...
+            $ref: '#/components/schemas/Base',
+          },
+        },
+        required: ['base'],
+        type: 'object',
+      },
+    });
+  });
+
+  it('treats objects created by .omit and .pick as new objects', () => {
```
AGalabov, line 182: I am not a fan of tests that check two things (i.e .omit AND .pick). Can you please separate them into separate it cases.

src/zod-extensions.ts
```diff
@@ -85,4 +85,20 @@ export function extendZodWithOpenApi(zod: typeof z) {
 
     return result;
   };
+
+  const zodPick = zod.ZodObject.prototype.pick as any;
+  zod.ZodObject.prototype.pick = function (this: any, ...args: any[]) {
+    const result = zodPick.apply(this, args);
+    result._def.openapi = {};
```
AGalabov, line 92: Maybe even `undefined` :thinking:? Same goes for omit.

AGalabov (commented):
Overall this sounds like the correct approach to clear the `_def.openapi`. Left minor comments, thanks again. I'll release a new 2.1.0 version once this gets merged

AGalabov (approved):

PR #42, opened 2022-10-09
Support zod preprocessed schemas

nktpro:
Since an OpenAPI request's underlying `params` and `query` values can only be `string`, it's very common having to resort to `z.preprocess(...)` to convert them to the desired type.

This PR adds support for `z.preprocess(...)`, similarly to the recent addition of support for `refine(...)`.

### DIFF
--- a/src/openapi-generator.ts
+++ b/src/openapi-generator.ts
@@ -584,7 +584,7 @@ export class OpenAPIGenerator {
 
     if (
       isZodType(zodSchema, 'ZodEffects') &&
-      zodSchema._def.effect.type === 'refinement'
+      (zodSchema._def.effect.type === 'refinement' || zodSchema._def.effect.type === 'preprocess')
     ) {
       const innerSchema = zodSchema._def.schema as ZodSchema<any>;
       return this.generateInnerSchema(innerSchema);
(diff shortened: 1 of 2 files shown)


src/openapi-generator.ts
```diff
@@ -584,7 +584,7 @@ export class OpenAPIGenerator {
 
     if (
       isZodType(zodSchema, 'ZodEffects') &&
-      zodSchema._def.effect.type === 'refinement'
+      (zodSchema._def.effect.type === 'refinement' || zodSchema._def.effect.type === 'preprocess')
```
AGalabov, line 587: I would imagine the build checks would fail because of prettier. This should be on two lines

nktpro: Thanks; updated after running `npm run prettier`

samchungy:
Just ran into this today, nice one!

heivo:
> Just ran into this today, nice one!

Same here, nice timing 😄

AGalabov (approved):

heivo:
@AGalabov would it be possible to release a new version soon that includes this change? thanks in advance!

AGalabov:
Yes, of course. Just drafted one. Should be done in a minute. v2.2.0 is the version

heivo:
Thank you very much!

PR #44, opened 2022-10-10
Add webhooks support

samchungy:
Hola! :) 

Open API 3.1 added `webhooks` which allow you to define event schema using the same route path objects that you can define paths with.
https://github.com/OAI/OpenAPI-Specification/blob/main/examples/v3.1/webhook-example.yaml

Adding support for Webhooks. I've also tested this on my current project and it produces schema which I validated successfully against the Open API 3.1 schema.

### DIFF
--- a/spec/routes.spec.ts
+++ b/spec/routes.spec.ts
@@ -31,474 +31,473 @@ const testDocConfig = {
   servers: [{ url: 'v1' }],
 };
 
-describe.each`
-  type          | registerFunction     | rootDocPath
-  ${'Routes'}   | ${'registerPath'}    | ${'paths'}
-  ${'Webhooks'} | ${'registerWebhook'} | ${'webhooks'}
-`(
-  '$type',
-  ({
-    registerFunction,
-    rootDocPath,
-  }: {
-    registerFunction: 'registerPath' | 'registerWebhook';
-    rootDocPath: 'paths' | 'webhooks';
-  }) => {
-    describe('response definitions', () => {
-      it('can set description', () => {
-        const registry = new OpenAPIRegistry();
+const routeTests = ({
+  registerFunction,
+  rootDocPath,
+}: {
+  registerFunction: 'registerPath' | 'registerWebhook';
+  rootDocPath: 'paths' | 'webhooks';
+}) => {
+  describe('response definitions', () => {
+    it('can set description', () => {
+      const registry = new OpenAPIRegistry();
 
-        registry[registerFunction]({
-          method: 'get',
-          path: '/',
-          responses: {
-            200: {
-              description: 'Simple response',
-              content: {
-                'application/json': {
-                  schema: z.string(),
-                },
-              },
-            },
-
-            404: {
-              description: 'Missing object',
-              content: {
-                'application/json': {
-                  schema: z.string(),
-                },
+      registry[registerFunction]({
+        method: 'get',
+        path: '/',
+        responses: {
+          200: {
+            description: 'Simple response',
+            content: {
+              'application/json': {
+                schema: z.string(),
               },
             },
           },
-        });
 
-        const document = new OpenAPIGenerator(
-          registry.definitions
-        ).generateDocument(testDocConfig);
-        const responses = document[rootDocPath]['/'].get.responses;
-
-        expect(responses['200'].description).toEqual('Simple response');
-        expect(responses['404'].description).toEqual('Missing object');
+          404: {
+            description: 'Missing object',
+            content: {
+              'application/json': {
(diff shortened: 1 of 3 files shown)


spec/routes.spec.ts
```diff
@@ -31,459 +31,474 @@ const testDocConfig = {
...
-      registry.registerPath({
-        method: 'get',
-        path: '/',
-        responses: {
-          200: {
-            description: 'Simple response',
-            content: {
-              'application/json': {
-                schema: z.string(),
+describe.each`
```
samchungy, line 34: No test changes here. Just running the same tests we run against routes against webhooks

spec/routes.spec.ts
```diff
@@ -31,459 +31,474 @@ const testDocConfig = {
...
+  ${'Webhooks'} | ${'registerWebhook'} | ${'webhooks'}
+`(
+  '$type',
+  ({
+    registerFunction,
+    rootDocPath,
+  }: {
+    registerFunction: 'registerPath' | 'registerWebhook';
+    rootDocPath: 'paths' | 'webhooks';
+  }) => {
```
georgyangelov, line 46: Can you please extract this function into a named one? That should make it a bit easier to read & hopefully there would be no indentation change needed and the diff would not be artificially big.

src/openapi-generator.ts
```diff
@@ -92,6 +93,10 @@ export class OpenAPIGenerator {
       ...config,
       components: this.buildComponents(),
       paths: this.pathRefs,
+      // As this is invalid in Open API 3.0 we optionally set it
```
georgyangelov, line 96: Good catch

src/openapi-generator.ts
```diff
@@ -151,7 +157,11 @@ export class OpenAPIGenerator {
         return;
 
       case 'route':
-        this.generateSingleRoute(definition.route);
+        this.generateSingleRoute(definition.type, definition.route);
+        return;
+
+      case 'webhook':
+        this.generateSingleRoute(definition.type, definition.webhook);
         return;
```
georgyangelov, line 165: You can unify the two cases:

```
case 'route':
case 'webhook':
  this.generateSingleRoute(definition.type, definition.webhook);
  return;
```

src/openapi-generator.ts
```diff
@@ -473,10 +483,19 @@ export class OpenAPIGenerator {
...
-      ...routeDoc,
-    };
+    if (type === 'route') {
+      this.pathRefs[path] = {
+        ...this.pathRefs[path],
+        ...routeDoc,
+      };
+    }
+
+    if (type === 'webhook') {
```
georgyangelov, line 493: How do you feel about a switch/case on this without a `default` case so that TS can verify if it's exhaustive or not? Though in this case it won't complain in both cases but still feels better.

georgyangelov (approved):
Very cool! 

Left a couple of cosmetic comments

src/openapi-generator.ts
AGalabov: Hmm instead of this I see that openapi3-ts has a newer major version - 3.1.1. And looking at [this line](https://github.com/metadevpro/openapi3-ts/blob/master/src/model/OpenApi.ts#L32) I think just upgrading the library would be a better shout.

src/openapi-generator.ts
```diff
@@ -473,10 +483,19 @@ export class OpenAPIGenerator {
       },
     };
 
-    this.pathRefs[path] = {
-      ...this.pathRefs[path],
-      ...routeDoc,
-    };
+    if (type === 'route') {
```
AGalabov, line 486: Hmm personally I would separate this into two methods - `generateSingleRoute` and `generateSingleWebhook`. The reusable code can be something like `generateEndpoint` or `generatePath` :thinking:? How do you feel about that?

AGalabov (changes requested):

src/openapi-generator.ts
AGalabov: Update I realized what the meaning of the optional setting is here. I also saw that you introduced the change there :smile:  I think it would be a good addition for you to update the dependency here as well :smile:

AGalabov (approved):

src/openapi-generator.ts
samchungy: Yeah I'm having a few existing type issues upgrading the library. Might raise a PR to separately to update it.

samchungy:
btw: I see you guys are Merging your pull requests in this repo, I'm normally used to squashing PRs to reduce commit spam but I'm happy to rebase commits if you'd like a cleaner commit history for your master.

PR #46, opened 2022-10-11
Enum support refactor suggestion

georgyangelov:
Thought to try it out instead of commenting. Fixes the issue of having mixed enums getting defined incorrectly as strings. See this change:

Previous: `expect(values).toEqual(['42', 'test']);`, new: `expect(values).toEqual([42, 'test']);`
and also mixed enums should be an error from OpenAPI generation standpoint because we can't support both without doing some `anyOf` schenanigans which I didn't want to think about.


spec/simple.spec.ts
```diff
@@ -199,6 +199,63 @@ describe('Simple', () => {
...
+    expectSchema([nativeEnumSchema], {
+      NativeEnum: {
+        type: 'string',
+        description: 'A native string enum in zod',
+        enum: ['a', 'b', 'c'],
+      },
+    });
+  });
+
+  it('does not support mixed native enums', () => {
```
AGalabov, line 223: Maybe a describe for enums could help here. Not sure if you already have this in the other PR

src/openapi-generator.ts
```diff
@@ -610,14 +611,26 @@ export class OpenAPIGenerator {
...
+
+      if (type === 'mixed') {
+        // enum Test {
+        //   A = 42,
+        //   B = 'test',
+        // }
+        //
+        // const result = z.nativeEnum(Test).parse('42');
+        //
+        // This is an error, so we can't just say it's a 'string'
```
AGalabov, line 624: Nice

src/lib/enum-info.ts
```diff
@@ -0,0 +1,24 @@
...
+  const numericCount = values.filter(_ => typeof _ === 'number').length;
+
+  const type =
+    numericCount === 0
+      ? ('string' as const)
+      : numericCount === values.length
+      ? ('numeric' as const)
+      : ('mixed' as const);
+
+  return { values: values, type };
```
AGalabov, line 23: Just `{ values, type }`

spec/simple.spec.ts
```diff
@@ -199,6 +199,63 @@ describe('Simple', () => {
     });
   });
 
+  it('supports native string enums', () => {
```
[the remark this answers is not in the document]
georgyangelov: How did I not see that

spec/simple.spec.ts
georgyangelov: I have that in the other PR

src/lib/enum-info.ts
georgyangelov: Done

AGalabov (approved):

PR #49, opened 2022-10-12
Update openapi3-ts to ^3.1.1

samchungy:


spec/types/object.spec.ts
```diff
@@ -146,7 +146,7 @@ describe('object', () => {
         });
 
       const DogSchema = AnimalSchema.extend({
-        type: z.string().openapi({ const: 'dog' }),
+        type: z.string().openapi({ example: 'dog' }),
```
samchungy, line 149: `const` is apparently not valid in open api so I've exchanged it with example here

src/openapi-generator.ts
```diff
@@ -102,15 +102,15 @@ export class OpenAPIGenerator {
     };
   }
 
-  generateComponents(): ComponentsObject {
```
samchungy, line 105: This type was incorrect

src/openapi-generator.ts
```diff
@@ -768,7 +773,10 @@ export class OpenAPIGenerator {
         );
       });
 
-      alreadyRequired = registeredSchema.required ?? [];
+      alreadyRequired =
+        'properties' in registeredSchema && registeredSchema.required
```
samchungy, line 777: Not as pretty, sorry 😢 but had to make the types happy

spec/routes.spec.ts
```diff
@@ -69,7 +69,7 @@ const routeTests = ({
       const document = new OpenAPIGenerator(
         registry.definitions
       ).generateDocument(testDocConfig);
-      const responses = document[rootDocPath]['/'].get.responses;
+      const responses = document[rootDocPath]?.['/'].get.responses;
```
samchungy, line 72: Since "webhooks" is now optional and no longer any we can either bang it or optional coerce. Doesn't really make a difference to me as it's just a test.

georgyangelov: This is fine, at least we won't get "cannot read whatever of undefined" but as you said, it's a test

src/openapi-generator.ts
georgyangelov: Nice catch

src/openapi-generator.ts
```diff
@@ -397,7 +397,7 @@ export class OpenAPIGenerator {
   ): SchemaObject | ReferenceObject {
     const simpleSchema = this.generateSimpleSchema(zodSchema);
 
-    if (simpleSchema.$ref) {
+    if ('$ref' in simpleSchema) {
```
georgyangelov, line 400: This is not exactly equivalent, maybe `if ('$ref' in simpleSchema && simpleSchema.$ref) {`?

src/openapi-generator.ts
georgyangelov: I'm wondering if this highlights an issue. The `registeredSchema` can be a ReferenceObject, which would mean it can be `{ $ref: <some-third-schema> }` itself. @AGalabov do you think we'll break in this case? So two extends: `A.extend({ ... }).extend({ ... }`

georgyangelov: Tried to add such a test and it failed, but I'm not sure this is the reason - maybe if we fix that, then this logic would break?

Screenshot 2022-10-14 at 0 28 04
Screenshot 2022-10-14 at 0 29 40

src/openapi-generator.ts
samchungy: Oopsie good catch!

src/openapi-generator.ts
samchungy: I'm a little confused here with your test. Is the extended schema 2 meant to extend schema 1?

georgyangelov: Yes, BaseSchema < ExtendedSchema < Extended2Schema, but later in the OpenAPI doc it's output as Base < Extended and Base < Extended2, so the Extended2 case is wrong

georgyangelov: Ah, I see what you mean, let me check again

georgyangelov: This is the real output if Extended2Schema extends ExtendedSchema

Screenshot 2022-10-17 at 18 01 12

georgyangelov: In any case, that's not a problem with your PR and we can solve it separately, I'll see if I can open a PR soon

georgyangelov (approved):

PR #50, opened 2022-10-12
Add discriminator key to ZodDiscriminatedUnion schemas

samchungy:
- Adds the `discriminator` key to discriminated unions generated schema only when a user has registered the schemas.
  - Added `mapping` for convenience for the user.
- Change the type of discriminated union to `oneOf`.

https://swagger.io/docs/specification/data-models/inheritance-and-polymorphism/

image

### DIFF
--- a/src/openapi-generator.ts
+++ b/src/openapi-generator.ts
@@ -18,6 +18,7 @@ import {
   DiscriminatorObject,
 } from 'openapi3-ts';
 import type {
+  map,
   ZodObject,
   ZodRawShape,
   ZodSchema,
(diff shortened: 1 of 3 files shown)


samchungy:
Looks like I need to do some mapping. Will get to it tomorrow

src/openapi-generator.ts
```diff
@@ -568,6 +569,35 @@ export class OpenAPIGenerator {
...
+  private mapDiscriminator(
+    zodObjects: ZodObject<any>[],
+    discriminator: string
+  ): DiscriminatorObject | undefined {
+    // All schemas must be registered to use a discriminator
+    if (zodObjects.some(obj => obj._def.openapi?.refId === undefined)) {
+      return undefined;
+    }
+
+    const mapping = zodObjects.reduce((map, obj) => {
```
georgyangelov, line 581: We're already mutating the map so we can simplify the code and use a `for` or `forEach` with a `const map: Record<string, string> = {}` at the top

georgyangelov (approved):
Very nice 🙇

src/openapi-generator.ts
samchungy: Gotcha 👍 it's always a style choice but I'll switch it over 🙂

Issue #48, opened 2022-10-12
Question: support ZodDate type


heivo:
Hi,

I'm using a preprocessor to cast ISO 8601 date string query params to Date objects.

Example code: `/api/v1/something?fromDate=2022-10-12T06%3A44%3A10Z`

My query definition is something like this:
```ts
z.object({
  fromDate: z
    .preprocess((arg) => {
      if (typeof arg === 'string') {
        const date = new Date(arg);
        if (isValid(date)) {
          return date;
        }
      }
      return arg;
    }, z.date())
    .optional()
    .openapi({
      type: 'string',
      example: new Date(),
      param: { description: 'From date filter, expects an ISO 8601 date string' },
    }),
})
```
It would be nice if the library would automatically handle this type properly so I wouldn't have to set the type field manually. The same also applies to request bodies that could contain serialized date strings.

If this is something you want to support I could contribute a PR, just let me know.

heivo:
Swagger UI renders the above example as
![image](https://user-images.githubusercontent.com/12901858/195272209-6d80e4cd-558a-4050-8e86-b7c5bd808ae9.png)

georgyangelov:
Hey @heivo , that's definitely something we want to support. It's essentially supporting the ZodDate type. Would be great if you can open a PR for that. Some pointers:

- You can add a test in `spec/types/date.spec.ts` similar to the other tests, you can run them with `npm run test`
- The main function that infers types is `OpenAPIGenerator#toOpenAPISchema`, you can add a check for ZodDate there. You will need to add the type to the `isZodType` function definition as well
- Make sure to run `npm run prettier` before submitting the PR - that will format the code according to the repo rules

Let me know if I can assist with anything!

AGalabov:
Fixed in #51

PR #51, opened 2022-10-14
feat: support ZodDate type

heivo:
#48

### DIFF
--- a/spec/types/date.spec.ts
+++ b/spec/types/date.spec.ts
@@ -1,0 +1,16 @@
+import { z } from 'zod';
+import { expectSchema } from '../lib/helpers';
+
+describe('date', () => {
+  it('supports dates and sets the type to "string"', () => {
+    const example = new Date('2022-10-14T09:39:30Z');
+    const schema = z.date().openapi({ refId: 'Date', example });
+
+    expectSchema([schema], {
+      Date: {
+        type: 'string',
+        example,
+      },
+    });
+  });
+});
--- a/spec/types/date.spec.ts
+++ b/spec/types/date.spec.ts
@@ -17,7 +17,7 @@ describe('date', () => {
     expectSchema([schema], {
       Date: {
         type: 'string',
-        example: example,
+        example,
       },
     });
   });
--- a/spec/types/date.spec.ts
+++ b/spec/types/date.spec.ts
@@ -2,14 +2,22 @@ import { z } from 'zod';
 import { expectSchema } from '../lib/helpers';
 
 describe('date', () => {
-  it('supports dates by setting the type to "string" and serializing examples', () => {
-    const example = new Date();
-    const schema = z.date().openapi({ refId: 'Date', example });
-
+  it('supports ZodDate and sets the type to `string`', () => {
+    const schema = z.date().openapi({ refId: 'Date' });
     expectSchema([schema], {
       Date: {
         type: 'string',
-        example: example.toISOString(),
+      },
+    });
+  });
+
+  it('uses `string` as the example type when the schema infers to `Date`', () => {
+    const example = new Date().toISOString();
+    const schema = z.date().openapi({ refId: 'Date', example });
+    expectSchema([schema], {
+      Date: {
+        type: 'string',
+        example: example,
       },
     });
   });
--- a/spec/types/date.spec.ts
+++ b/spec/types/date.spec.ts
@@ -2,14 +2,14 @@ import { z } from 'zod';
 import { expectSchema } from '../lib/helpers';
 
 describe('date', () => {
-  it('supports dates and sets the type to "string"', () => {
-    const example = new Date('2022-10-14T09:39:30Z');
+  it('supports dates by setting the type to "string" and serializing examples', () => {
+    const example = new Date();
     const schema = z.date().openapi({ refId: 'Date', example });
 
     expectSchema([schema], {
       Date: {
         type: 'string',
-        example,
+        example: example.toISOString(),
       },
     });
   });
--- a/src/openapi-generator.ts
+++ b/src/openapi-generator.ts
@@ -711,11 +711,9 @@ export class OpenAPIGenerator {
     }
 
     if (isZodType(zodSchema, 'ZodDate')) {
-      const metadata = this.getMetadata(zodSchema);
       return {
         type: 'string',
         nullable: isNullable ? true : undefined,
-        example: metadata?.example?.toISOString(),
       };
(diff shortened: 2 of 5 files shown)


spec/types/date.spec.ts
```diff
@@ -0,0 +1,16 @@
...
+
+describe('date', () => {
+  it('supports dates and sets the type to "string"', () => {
+    const example = new Date('2022-10-14T09:39:30Z');
+    const schema = z.date().openapi({ refId: 'Date', example });
+
+    expectSchema([schema], {
+      Date: {
+        type: 'string',
+        example,
```
georgyangelov, line 12: Since it's a type: string the example should probably be `example: example.toISOString()` instead of the date object itself. This way we're not relying on the yaml or json serializer to decide how to display it for us.

georgyangelov (commented):
Nice, thank you!

spec/types/date.spec.ts
heivo: Thank you for the suggestion, I'm now serializing examples in the generator to ISO date strings. Is this what you meant?

heivo: Maybe this is too opinionated because not everyone might want to use ISO date strings. Instead we could change the example TS type from Date to string when the schema is a ZodDate (or derive the example type from the openapi metadata type). What do you think?

src/openapi-generator.ts
```diff
@@ -868,8 +879,8 @@ export class OpenAPIGenerator {
   ): SchemaObject | ReferenceObject {
     return omitBy(
       {
-        ...initialData,
         ...this.buildSchemaMetadata(metadata),
+        ...initialData,
```
AGalabov, line 883: Hmmm what caused this change? I think that the metadata should override anything coming as a base schema.

src/openapi-generator.ts
```diff
@@ -853,7 +862,9 @@ export class OpenAPIGenerator {
     return omitBy(metadata, isNil);
   }
 
-  private getMetadata(zodSchema: ZodSchema<any>) {
+  private getMetadata<T extends any>(
```
AGalabov, line 865: Does inferring the type help with anything here? I am okay with this change, just asking if it is cosmetic or if it helped with a TS usecase.

src/openapi-generator.ts
```diff
@@ -710,6 +710,15 @@ export class OpenAPIGenerator {
       return {};
     }
 
+    if (isZodType(zodSchema, 'ZodDate')) {
+      const metadata = this.getMetadata(zodSchema);
+      return {
+        type: 'string',
+        nullable: isNullable ? true : undefined,
+        example: metadata?.example?.toISOString(),
```
AGalabov, line 718: Seeing that you are going with the `toISOString` approach then we could have added a `date-time` `format`. However, looking at [this](https://swagger.io/docs/specification/data-models/data-types/) I see that there are both `date` and `date-time`. This means that for a "user" of `zod-to-openapi` it might be a perfectly valid use-case to use a `date` format. And since the examples are being set by the "user" I'd rather leave this to them to serialize and not touch it at all here.

spec/types/date.spec.ts
AGalabov: I think that the example should be defined by the user and not zod-to-openapi. But yes, it should be a string (Typescript wise)

src/openapi-generator.ts
heivo: I added it to have proper types here, otherwise the example would be `any` type.
image

I will probably not need it once I updated the PR but I think it's a good addition nevertheless. But I can also revert it if you want to.

spec/types/date.spec.ts
heivo: Ok I changed the example type to `string` when the schema type infers to `Date`. I also reverted the automatically generated example so users can define them.

src/zod-extensions.ts
```diff
@@ -5,8 +5,10 @@ import { isZodType } from './lib/zod-is-type';
 export interface ZodOpenAPIMetadata<T = any> extends SchemaObject {
   refId?: string;
   extendedFrom?: string;
-  param?: Partial<ParameterObject> & { example?: T };
-  example?: T;
+  param?: Partial<ParameterObject> & {
+    example?: T extends Date ? string : T;
```
AGalabov, line 9: @heivo can you please pull the latest master. Since in [this PR](https://github.com/asteasolutions/zod-to-openapi/pull/53/files) the same typing has been used for `examples` and `default`. You can also create a reusable type to remove the duplication.

spec/types/date.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+    });
+  });
+
+  it('uses `string` as the example type when the schema infers to `Date`', () => {
+    const example = new Date().toISOString();
+    const schema = z.date().openapi({ refId: 'Date', example });
+    expectSchema([schema], {
+      Date: {
+        type: 'string',
+        example: example,
```
AGalabov, line 20: No need for `example: example` - they share the same name

AGalabov (approved):

AGalabov:
@heivo The PR looks good. However the build checks are failing - presumably because of prettier. Please run `npm run lint` (it checks for errors) and `npm run prettier` (fixes them automatically) and finally `npm test` to verify everything is working fine. I assume there are styling changes that should be made. After that I'll merge your PR

heivo:
> @heivo The PR looks good. However the build checks are failing - presumably because of prettier. Please run `npm run lint` (it checks for errors) and `npm run prettier` (fixes them automatically) and finally `npm test` to verify everything is working fine. I assume there are styling changes that should be made. After that I'll merge your PR

The build fails because of a type error but I think it's not related to my changes, the same happens for me in master branch as well

AGalabov:
@heivo you are right. It must have happened when the latest PRs have been merged (especially the one for the upgrade of `openapi3-ts`). I opened a [PR](https://github.com/asteasolutions/zod-to-openapi/pull/56) with the fix that seemed best suited to me. I'll check with @georgyangelov if we can agree on it. Sorry for the inconvenience. Will let you know once a fix is merged.

AGalabov:
Well that was quick :smile: .... @heivo we merged the PR so you should be able to fix the issues once you pull master into your branch.

georgyangelov (approved):

heivo:
> Well that was quick smile .... @heivo we merged the PR so you should be able to fix the issues once you pull master into your branch.

@AGalabov I merged the master branch

PR #52, opened 2022-10-15
fix: ignore existing metadata when referencing registered schema

samchungy:
I currently have a small issue where existing metadata on a registered schema gets duplicated when it is referenced. This resolves that issue. In my example budget already contains that description and I never declared it again

Before/After:
image

### DIFF
--- a/src/openapi-generator.ts
+++ b/src/openapi-generator.ts
@@ -363,18 +363,22 @@ export class OpenAPIGenerator {
         $ref: `#/components/schemas/${refId}`,
       };
 
-      const nullableMetadata =
-        !schemaRef['nullable'] && zodSchema.isNullable()
-          ? { nullable: true }
-          : {};
-
-      const matchingValueKeys = Object.keys(metadata).filter(key =>
-        objectEquals(metadata[key], schemaRef[key])
+      // New metadata from ZodObject properties.
+      const newSchemaMetadata = omitBy(
+        this.toOpenAPISchema(innerSchema, zodSchema.isNullable()),
+        (value, key) =>
+          value === undefined || objectEquals(value, schemaRef[key])
       );
-      const newMetadata = omit(metadata, matchingValueKeys);
+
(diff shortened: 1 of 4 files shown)


src/openapi-generator.ts
```diff
@@ -353,22 +353,29 @@ export class OpenAPIGenerator {
...
       };
 
-      const nullableMetadata = zodSchema.isNullable() ? { nullable: true } : {};
+      const nullableMetadata =
+        !schemaRef['nullable'] && zodSchema.isNullable()
+          ? { nullable: true }
+          : {};
+
+      const matchingValueKeys = Object.keys(metadata).filter(
+        key => metadata[key] === schemaRef[key]
```
samchungy, line 372: Hmm just realised this might not work as the value could be something other than a string. Will think about it

src/openapi-generator.ts
```diff
@@ -353,22 +353,33 @@ export class OpenAPIGenerator {
...
-      const referenceObject = {
+      const schemaRef = this.schemaRefs[refId] as SchemaObject;
+      const referenceObject: ReferenceObject = {
         $ref: `#/components/schemas/${refId}`,
       };
 
-      const nullableMetadata = zodSchema.isNullable() ? { nullable: true } : {};
+      // New metadata from ZodSchema properties.
+      const newSchemaMetadata = omitBy(
+        this.toOpenAPISchema(innerSchema, zodSchema.isNullable()),
```
AGalabov, line 368: In this same function we have a check if there is a provided `type` in the metadata and skip the generation logic in such case. I feel like the same should be applicable here. Basically in my opinion if a user opts to provide a `type` we should stop trying to infer properties since they might not be correct for his case.

For example `z.string().regex(/\d*/).openapi({ type: 'number' })` - the generated response should be `{ type: 'number' }` and not `{ type: 'number', pattern: '\d*' }`.

AGalabov (approved):

PR #53, opened 2022-10-15
Add typing for `examples` and `default` metadata

samchungy:
Should be a simple one. Just add some typing for `examples` since `example` is invalid in Open API 3.1. Also added a type for `default`. I just realised that you can use `.default()` but anyway better typing is always nice.

Updated some unit tests examples to prove the types work.


AGalabov (approved):
Looks good. However, I would've separated the tests. A nice example in my opinion would be in `/modifieres/default` where you can verify that a manually added `default` in `.openapi` takes precendence to one generated from `zod`.

PR #54, opened 2022-10-16
Fix `required` logic for parameters

samchungy:
Use the existing `isOptionalSchema` function to calculate if path parameters are optional.


georgyangelov (approved):
Cool :)

PR #55, opened 2022-10-17
Fix object inheritance when it's more than one level deep

georgyangelov:


src/zod-extensions.ts
```diff
@@ -48,13 +48,15 @@ export function extendZodWithOpenApi(zod: typeof z) {
...
+      const originalExtend = this.extend;
 
       result.extend = function (...args: any) {
-        const extendedResult = initialExtend.apply(result, args);
+        const extendedResult = originalExtend.apply(this, args);
 
         extendedResult._def.openapi = {
-          extendedFrom: result._def.openapi?.refId,
+          extendedFrom: this._def.openapi?.refId
+            ? { refId: this._def.openapi?.refId, schema: this }
```
AGalabov, line 58: You can remove the optional chaining here, right?

src/zod-extensions.ts
```diff
@@ -48,13 +48,15 @@ export function extendZodWithOpenApi(zod: typeof z) {
     });
 
     if (isZodType(this, 'ZodObject')) {
-      const initialExtend = this.extend;
+      const originalExtend = this.extend;
 
       result.extend = function (...args: any) {
-        const extendedResult = initialExtend.apply(result, args);
+        const extendedResult = originalExtend.apply(this, args);
```
AGalabov, line 54: Hmmm I can't remember why I did this. Can you explain what the difference is? What was the unexpected behavior beforehand?

spec/types/object.spec.ts
```diff
@@ -93,6 +93,110 @@ describe('object', () => {
...
+                  type: 'number',
+                },
+              },
+            },
+          ],
+        },
+      });
+    });
+
+    it('can chain-extend objects correctly without intermediate link', () => {
```
AGalabov, line 154: This file is once again getting too big. I think it makes sense to move the tests for `.extend` in `modifiers/extends`. It is not exactly a modifier but it is pretty much there (we can change the name of the folder if need be).

AGalabov (approved):

src/zod-extensions.ts
georgyangelov: Should be pretty much the same. The difference is in this code:

```ts
const a = z.object({ a: 1 })
const b = z.object({ b: 1 })

b.extend = a.extend

b.extend({ c: 1 })
// Before: { a: 1, c: 1 }
// After: { b: 1, c: 1 }
```

Since the method was defined on the prototype, the second behavior is the more correct one since it was relying on `this` before and the method wasn't bound.

georgyangelov: Actually was it defined on the prototype hm, let me check

georgyangelov: Ok so it looks like the `originalExtend` is already bound so my change won't change anything. Still, in the other case - if it wasn't bound, then the original definition would change behavior while the new one won't.

src/openapi-generator.ts
```diff
@@ -783,59 +783,46 @@ export class OpenAPIGenerator {
...
-      this.generateInnerSchema(propSchema)
-    );
-
-    let alreadyRegistered: string[] = [];
-    let alreadyRequired: string[] = [];
-
-    if (extendedFrom) {
-      const registeredSchema = this.schemaRefs[extendedFrom];
-
-      if (!registeredSchema) {
```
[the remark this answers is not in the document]
georgyangelov: We don't need it because we now only have `extendedFrom` if we're extending from a defined schema. And we're not using that schema from `schemaRefs` so we don't need to double-check.

src/zod-extensions.ts
georgyangelov: Done

PR #57, opened 2022-10-19
Open API 3.1.0 Support

samchungy:
# BREAKING CHANGE

The generator's constructor now takes a mandatory Open API version. This is required in order to drive different generation behaviour based on the version.

`generateDocument` no longer takes a `openapi` version and instead we will use the version from the generator's constructor to populate this field.

## Why we need this change

Open API 3.1.0 made `nullable` an invalid key and instead chose to become JSON Schema compatible by using a type array with `null` inside it. The problem here is that the `type` array is invalid in previous versions.

https://www.openapis.org/blog/2021/02/16/migrating-from-openapi-3-0-to-3-1-0

image

## Other Changes

Also adds nullable support for:
- Union
- Discriminated Union
- Intersection
- Array
- Record

### DIFF
--- a/spec/types/intersection.spec.ts
+++ b/spec/types/intersection.spec.ts
@@ -31,4 +31,39 @@ describe('intersection', () => {
       }
     );
   });
+
+  it('supports nullable intersection types', () => {
+    const Person = z.object({
+      name: z.string(),
+    });
+
+    const Employee = z.object({
+      role: z.string(),
+    });
+
+    expectSchema(
+      [z.intersection(Person, Employee).nullable().openapi({ refId: 'Test' })],
+      {
+        Test: {
+          anyOf: [
+            {
+              allOf: [
+                {
+                  type: 'object',
+                  properties: { name: { type: 'string' } },
+                  required: ['name'],
+                },
+                {
+                  type: 'object',
+                  properties: { role: { type: 'string' } },
+                  required: ['role'],
+                },
+              ],
+            },
+            { nullable: true },
+          ],
+        },
+      }
+    );
+  });
 });
--- a/spec/types/union.spec.ts
+++ b/spec/types/union.spec.ts
@@ -39,4 +39,16 @@ describe('union', () => {
       }
     );
   });
+
+  it('supports nullable union types in 3.1.0', () => {
+    expectSchema(
+      [z.string().or(z.number()).nullable().openapi({ refId: 'Test' })],
+      {
+        Test: {
+          anyOf: [{ type: 'string' }, { type: 'number' }, { type: 'null' }],
+        },
+      },
+      '3.1.0'
+    );
+  });
 });
--- a/spec/types/union.spec.ts
+++ b/spec/types/union.spec.ts
@@ -28,4 +28,15 @@ describe('union', () => {
       }
     );
   });
+
+  it('supports nullable union types', () => {
+    expectSchema(
+      [z.string().or(z.number()).nullable().openapi({ refId: 'Test' })],
+      {
+        Test: {
+          anyOf: [{ type: 'string' }, { type: 'number' }, { nullable: true }],
+        },
+      }
+    );
+  });
 });
--- a/src/openapi-generator.ts
+++ b/src/openapi-generator.ts
@@ -61,7 +61,7 @@ export type OpenApiVersion = typeof openApiVersions[number];
 // Omit does not work, since OpenAPIObject extends ISpecificationExtension
 // and is inferred as { [key: number]: any; [key: string]: any }
 export interface OpenAPIObjectConfig {
-  openapi: OpenApiVersion;
+  openapi?: OpenApiVersion;
   info: InfoObject;
   servers?: ServerObject[];
   security?: SecurityRequirementObject[];
@@ -87,18 +87,23 @@ export class OpenAPIGenerator {
     name: string;
     component: OpenAPIComponentObject;
   }[] = [];
-  private openAPIVersion: OpenApiVersion = '3.0.0';
 
-  constructor(private definitions: OpenAPIDefinitions[]) {
+  constructor(
+    private definitions: OpenAPIDefinitions[],
+    private openAPIVersion: OpenApiVersion
+  ) {
     this.sortDefinitions();
   }
 
   generateDocument(config: OpenAPIObjectConfig): OpenAPIObject {
-    this.openAPIVersion = config.openapi;
+    if (config.openapi) {
+      this.openAPIVersion = config.openapi;
+    }
     this.definitions.forEach(definition => this.generateSingle(definition));
 
     return {
       ...config,
+      openapi: this.openAPIVersion,
       components: this.buildComponents(),
       paths: this.pathRefs,
       // As the `webhooks` key is invalid in Open API 3.0.x we need to optionally set it
@@ -111,7 +116,9 @@ export class OpenAPIGenerator {
   generateComponents(
     config?: Pick<OpenAPIObjectConfig, 'openapi'>
   ): Pick<OpenAPIObject, 'components'> {
-    this.openAPIVersion = config?.openapi ?? this.openAPIVersion;
+    if (config?.openapi) {
+      this.openAPIVersion = config.openapi;
+    }
 
     this.definitions.forEach(definition => this.generateSingle(definition));
 
--- a/src/openapi-generator.ts
+++ b/src/openapi-generator.ts
@@ -816,12 +816,7 @@ export class OpenAPIGenerator {
 
       if (isNullable) {
         return {
-          anyOf: [
-            allOfSchema,
-            ...(isNullable
-              ? [this.mapNullableType(undefined, isNullable)]
-              : []),
-          ],
+          anyOf: [allOfSchema, this.mapNullableType(undefined, isNullable)],
         };
       }
 
--- a/src/openapi-generator.ts
+++ b/src/openapi-generator.ts
@@ -90,7 +90,7 @@ export class OpenAPIGenerator {
   }
 
   generateDocument(config: OpenAPIObjectConfig): OpenAPIObject {
-    this.openAPIVersion = config.openapi; // Preference the constructor version over this one
+    this.openAPIVersion = config.openapi;
     this.definitions.forEach(definition => this.generateSingle(definition));
 
     return {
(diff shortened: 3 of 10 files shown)


src/openapi-generator.ts
```diff
@@ -84,12 +83,14 @@ export class OpenAPIGenerator {
     name: string;
     component: OpenAPIComponentObject;
   }[] = [];
+  private openAPIVersion: string | undefined;
```
AGalabov, line 86: Hmmm a few things here.
1. How do you feel about making this an enum instead of a plain string. That way we would be the ones defining the possible/suported OpenAPI versions.
2. I would make this required and also a part of the `OpenAPIGenerator` constructor instead of using it upon the method invocation.
3. For `generateDocument` we can make `openapi` optional - if it is not provided we would use the value from the constructor. If it is provided - we would use it. That way the generation logic would be explicit and at the same time we would leave an option for the user to spcify the version at his own risk.

src/openapi-generator.ts
```diff
@@ -620,6 +625,23 @@ export class OpenAPIGenerator {
     };
   }
 
+  private mapNullableType(
+    type: NonNullable<SchemaObject['type']>,
+    isNullable: boolean
+  ): Pick<SchemaObject, 'type' | 'nullable'> {
+    // Open API 3.1.0 made the `nullable` key invalid and instead you use type arrays
+    if (isNullable && this.openAPIVersion && this.openAPIVersion >= '3.1.0') {
```
AGalabov, line 633: Using an enum we would be able to specify exact behaviors. Comparing version strings should not be done like that. For example `'10.1.1' > '3.1.0'` would evaluate to `false` (obviously that's way ahead of time but it is an example). There is a [library](https://www.npmjs.com/package/compare-versions) for that but if we can implement it ourselves utilizing enums I think that would be better.

AGalabov (commented):
Thank you @samchungy for bringing this up.

I have a few questions that arose when I reviewed this.
1. We have `nullable` for SOME of out schemas, but there are also some that do not have it. For example `ZodDiscriminatedUnion`, `ZodUnion`, `ZodArray` and others. @georgyangelov do you remember if there was a particular reason for this behavior - I don't think so.
2. @samchungy do you happen to know if there are any other changes that occurred with v3.1 that we should take into consideration. I'd try and take a look myself during the weekend. The idea is that the next version would be a major one (due to the upgrade of `openapi3-ts` whose types we export). And it would be great if we can call this release "Support for OpenAPI 3.1" with a reasonable certainty that we've covered all of the updates.

src/openapi-generator.ts
samchungy: Ahhhh yeah stupid me 🫣

src/openapi-generator.ts
samchungy: afaik this is the main breaking change for open API 3.1.0. The other changes are listed in the article above. example -> examples consumers can change manually. I can have a look into the union/array nullables.

src/openapi-generator.ts
samchungy: I'm not sure if what I've done is what you had in mind, if we want to allow unspecified/newer versions I can always rip their `satisfies` code out into our repo 😄

src/openapi-generator.ts
```diff
@@ -54,11 +53,15 @@ import { enumInfo } from './lib/enum-info';
...
+const openApiVersions = ['3.0.0', '3.0.1', '3.0.2', '3.0.3', '3.1.0'] as const;
+
+export type OpenApiVersion = typeof openApiVersions[number];
+
 // This is essentially OpenAPIObject without the components and paths keys.
 // Omit does not work, since OpenAPIObject extends ISpecificationExtension
 // and is inferred as { [key: number]: any; [key: string]: any }
-interface OpenAPIObjectConfig {
-  openapi: string;
+export interface OpenAPIObjectConfig {
```
AGalabov, line 63: Hmmm now that you updated the library I believe we can use Omit here. Because the ISpecificationExtension now has a key type of `x-${string}`. Can you check.

src/openapi-generator.ts
```diff
@@ -85,15 +88,22 @@ export class OpenAPIGenerator {
     component: OpenAPIComponentObject;
   }[] = [];
 
-  constructor(private definitions: OpenAPIDefinitions[]) {
+  constructor(
+    private definitions: OpenAPIDefinitions[],
+    private openAPIVersion: OpenApiVersion = '3.0.0'
```
AGalabov, line 93: Since we are to make a major version with breaking changes. I feel like we can make this a mandatory field. How do you feel about that :thinking:

src/openapi-generator.ts
```diff
@@ -85,15 +88,22 @@ export class OpenAPIGenerator {
...
+  ) {
     this.sortDefinitions();
   }
 
   generateDocument(config: OpenAPIObjectConfig): OpenAPIObject {
+    if (config.openapi) {
+      this.openAPIVersion = config.openapi;
```
AGalabov, line 100: Hmm this seems both correct and incorrect.
1. I feel like we should not be modifying the value provided in the constructor. It's quite implicit. And I would be okay for the user to "break his head" at his own risk
2. I do get your point that we should try and "keep him from breaking his head"

My gut feeling is more inclined to go with 1. The best example of why I prefer it is something like the following:
```ts
const generator = new OpenAPIGenerator(schemas, '3.0.0');
generator.generateComponents(); // this would have output for 3.0.0
generator.generateDocument({ openapi: '3.1.0', ... });
generator.generateComponents(); // this would have output for 3.1.0
```
The reason for the change in behavior of `generateComponents` is not really obvious.
cc: @georgyangelov  any opinion?

src/openapi-generator.ts
```diff
@@ -103,7 +113,13 @@ export class OpenAPIGenerator {
     };
   }
 
-  generateComponents(): Pick<OpenAPIObject, 'components'> {
+  generateComponents(
+    config?: Pick<OpenAPIObjectConfig, 'openapi'>
```
AGalabov, line 117: I would personally remove this - same reasoning as above. If we do keep it - I feel like we should rather keep this as a simple schema - `{ openapi?: OpenApiVersion }` - no need to "link" it to the OpenAPIObjectConfig

src/openapi-generator.ts
```diff
@@ -103,7 +113,13 @@ export class OpenAPIGenerator {
...
+  ): Pick<OpenAPIObject, 'components'> {
+    if (config?.openapi) {
+      this.openAPIVersion = config.openapi;
+    }
+
     this.definitions.forEach(definition => this.generateSingle(definition));
```
AGalabov, line 123: Actually I just thought about this. The example above is not really correct. Right now if we generate schemas once and call (for example) `generateComponents` once again we would not be "generating" them since they would be kept in memory and we would only reuse the currrent refIds. So providing a different version value won't actually work when called consecutively.

src/openapi-generator.ts
```diff
@@ -729,32 +784,50 @@ export class OpenAPIGenerator {
...
       const options = [...zodSchema.options.values()];
 
       return {
-        oneOf: options.map(schema => this.generateInnerSchema(schema)),
-        discriminator: this.mapDiscriminator(options, zodSchema.discriminator),
+        oneOf: [
+          ...options.map(schema => this.generateInnerSchema(schema)),
+          ...(isNullable ? [this.mapNullableType(undefined, isNullable)] : []),
+        ],
+        discriminator: this.mapDiscriminator(
```
AGalabov, line 802: Maybe we can move the logic with `isNullable` here instead. I.e `isNullable ? this.mapDiscriminator(...) : undefined`. That way the comment "If schema is nullable we should treat it like a union." would make even more sense here.

src/openapi-generator.ts
```diff
@@ -620,6 +640,40 @@ export class OpenAPIGenerator {
...
+    openApiVersions.indexOf(comparison);
+
+  private mapNullableType(
+    type: SchemaObject['type'],
+    isNullable: boolean
+  ): Pick<SchemaObject, 'type' | 'nullable'> {
+    // Open API 3.1.0 made the `nullable` key invalid and instead you use type arrays
+    if (this.openApiVersionSatisfies(this.openAPIVersion, '3.1.0')) {
+      if (isNullable) {
+        if (type === undefined) {
```
AGalabov, line 657: Hmm this seems a little but obscure. Considering the usage is `...(isNullable ? [this.mapNullableType(undefined, isNullable)] : [])`. Basically we always call this function with `undefined` and `true`. Maybe it would be better to create a separate function for thosecases. It can either handle the whole ternary operator as well => returns an array itself. Or it can just handle "the "generation of the `nullable` parameter for oneOf/allOf/anyOf"

spec/types/intersection.spec.ts
```diff
@@ -31,4 +31,39 @@ describe('intersection', () => {
       }
     );
   });
+
+  it('supports nullable intersection types', () => {
```
AGalabov, line 35: Thank you! :bowing_man:  Those were not working before, were they?

spec/types/union.spec.ts
```diff
@@ -28,4 +28,15 @@ describe('union', () => {
       }
     );
   });
+
+  it('supports nullable union types', () => {
```
AGalabov, line 32: Since this one is handled in a different manner based on versions it makes sense to add one more test for 3.1 as well?

src/openapi-generator.ts
AGalabov: Yup using a simple array is what I was thinking of as well

src/openapi-generator.ts
samchungy: Yeah okay that makes sense to me. Wasn't sure if we wanted to go with breaking change or not but not mutating the constructor's version makes sense.

src/openapi-generator.ts
samchungy: I think I've made it clearer. I agree it was confusing to have isNullable driving behaviour in different areas.

spec/types/intersection.spec.ts
samchungy: They were not!

samchungy:
Question: should I add nullable tests for all the different types?

src/openapi-generator.ts
samchungy: Ahhhhh string template types. Should've thought of that 🧠

AGalabov:
@samchungy 
> Question: should I add nullable tests for all the different types?
Don't think that's necessary.

AGalabov (approved):

PR #59, opened 2022-10-23
supports minLength/maxLength on string type

wiz-kakiuchi:
**Note:** 
Number.isFinite" is used because "zod v3.19.1" has not yet been applied to "zod-to-openapi".

See below for details.
https://github.com/colinhacks/zod/commit/d78047e9f44596a96d637abb0ce209cd2732d88c

### DIFF
--- a/src/openapi-generator.ts
+++ b/src/openapi-generator.ts
@@ -632,6 +632,13 @@ export class OpenAPIGenerator {
       const regexCheck = this.getZodStringCheck(zodSchema, 'regex');
       return {
         type: 'string',
+        // FIXME: https://github.com/colinhacks/zod/commit/d78047e9f44596a96d637abb0ce209cd2732d88c
+        minLength: Number.isFinite(zodSchema.minLength)
+          ? zodSchema.minLength ?? undefined
+          : undefined,
+        maxLength: Number.isFinite(zodSchema.maxLength)
+          ? zodSchema.maxLength ?? undefined
+          : undefined,
         nullable: isNullable ? true : undefined,
         format: this.mapStringFormat(zodSchema),
         pattern: regexCheck?.regex.source,
--- a/spec/types/string.spec.ts
+++ b/spec/types/string.spec.ts
@@ -8,6 +8,15 @@ describe('string', () => {
     });
   });
 
+  it('supports minLength / maxLength on string', () => {
+    expectSchema(
+      [z.string().min(5).max(10).openapi({ refId: 'minMaxLengthString' })],
+      {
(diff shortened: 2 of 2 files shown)


src/openapi-generator.ts
```diff
@@ -632,6 +632,13 @@ export class OpenAPIGenerator {
       const regexCheck = this.getZodStringCheck(zodSchema, 'regex');
       return {
         type: 'string',
+        // FIXME: https://github.com/colinhacks/zod/commit/d78047e9f44596a96d637abb0ce209cd2732d88c
+        minLength: Number.isFinite(zodSchema.minLength)
+          ? zodSchema.minLength ?? undefined
```
AGalabov, line 637: I don't think you need the nullish coalescing (`??`) here, do you? If it is `null` (new versions) - the condition above would evaluate to false and we would return `undefined` anyway.

src/openapi-generator.ts
```diff
@@ -632,6 +632,13 @@ export class OpenAPIGenerator {
       const regexCheck = this.getZodStringCheck(zodSchema, 'regex');
       return {
         type: 'string',
+        // FIXME: https://github.com/colinhacks/zod/commit/d78047e9f44596a96d637abb0ce209cd2732d88c
```
AGalabov, line 635: Hmm that is interesting. Great catch. I am not sure if I want to raise the minimal zod version for this check only. So I would rather keep this in its current state for now (using `isFinite`). Once we get to update for a different reason we can revise this. cc: @georgyangelov any thoughts?

AGalabov (approved):

src/openapi-generator.ts
samchungy: You can probably potentially just bump up your dev deps and pass in `zodSchema.minLength` since it won't be defined in previous versions it'll end up as just undefined?

AGalabov: Uhmmm @samchungy  from what I saw both `minLength` and `maxLength` exist in the current dependency version (3.14.3). The only difference being the default that is being used - `null` for older versions and `Infinity` for the latest zod version. And seeing that we cannot pass an infinity to openapi3ts -> I think that `isFinite` is the correct check overall

AGalabov: @wiz-kakiuchi do you mind resolving the conflicts and possibly fixing the nullish coalescence above.

AGalabov:
@wiz-kakiuchi I wanted to merge this so I opened a new PR where I resolved the conflicts. I made sure to use your commit so that you still get credit for your work. Thank you again for contributing!

PR #61, opened 2022-10-24
Expose a metadata getter 

AGalabov:


src/openapi-metadata.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  /**
+   * A method that omits all custom keys added to the regular OpenAPI
+   * metadata properties
+   */
+  static extractOpenApiMetadata(metadata: ZodOpenAPIMetadata) {
+    return omitBy(omit(metadata, ['param', 'refId', 'extendedFrom']), isNil);
+  }
+
+  static unwrapChained(schema: ZodSchema<any>): ZodSchema<any> {
```
AGalabov, line 51: I am not really a fan of this method being moved here. I wanted to push this code to get some feedback + I need to do a software update, so I'd rather not keep this locally :smile:. Any alternative ideas are more than welcome on the whole structure.

georgyangelov:
What are we solving here?

src/openapi-metadata.ts
```diff
@@ -0,0 +1,69 @@
...
+  ): ZodOpenAPIMetadata<T> | undefined {
+    const innerSchema = this.unwrapChained(zodSchema);
+    const metadata = zodSchema._def.openapi
+      ? zodSchema._def.openapi
+      : innerSchema._def.openapi;
+
+    return metadata;
+  }
+
+  static applySchemaMetadata(
```
georgyangelov, line 25: I'm not a fan of the statics, can you please make it a normal object? We can instantiate it once in `OpenApiGenerator`.

samchungy:
Hmmm personally I'm torn with the `internal_openapi` method. On one hand I understand why you want to protect it but on the other hand I use it right now because I have a separation between my Zod Types, Registry and Generator. 

I actually register some of my types like this:

eg.

```ts
// types package
const myType = z.object({str: z.string()}).openapi({refId: 'myObject'})

// scripts/generate-schema.ts
const registry = new OpenAPIRegistry()
const registeredSchemas = [myType];
registeredSchemas.forEach((schema) => {
  registry.register(schema._def.openapi?.refId as string, schema);
});

// ... generator code
```
Just because it feels a little a little clunky to create the zod type, then refer to the registered version? 🤔 

```ts
const registry = new OpenAPIRegistry()

const myType = z.object({str: z.string()})
const registeredType =  registry.register('myObject', myType);

const newType = z.object({t: registeredType});
```

AGalabov:
@samchungy I am not sure I understand your last example. Why can't you just do:
```ts
// types package
//registry
export const registry = new OpenAPIRegistry()

// foo.ts
import {registry} from './registry';
const myType = registry.register('myObjec', tz.object( { str: z.string() } ))

const newType = z.object({t: myType});

//bar.ts
import {registry} from './registry';
const myType2 = registry.register('myObject2', z.object( { str: z.string() } ))
```

This is exactly what we do in out own project. In our case we do have a types package as well. Everything that needs to be registered is getting registered there. In the build script you just import the `registry` and pass it to the `generator`. I see no reason to separate the "schema" and the "registered schema".

Saying all that `internal_openapi` is "deprecated" but anyone can still use it ~at your own risk~ - it's just that we do not encourage you to do it since it might break something in the generation logic.

samchungy:
@AGalabov all good 😄 Just thought I could get away with not initialising and creating a register to register() within my types package. I ran into issues with my way later on anyway with needing to register things in order so I might just do it the register way from now on.

georgyangelov (approved):

samchungy:
There's no chat feature but that's a bunch for the release guys @AGalabov @georgyangelov Appreciate your work 🙌 🎉

AGalabov:
@samchungy  thank you for the kind words. And even more so - once again thank you for contributing. Quite a bit of the latest release was down to your work :bow:.

PR #63, opened 2022-10-28
Default Support

### LINKED ISSUE #60: Properly handle zod default
Currently we do have support for the `.default` modifier but we do not add the default value in the generated schema.

Dev note: This issue would require some refactoring as currently schemas are being unwrapped using `unwrapChained` before they are being sent to `toOpenAPISchema`. That also makes the if for `isZodType(zodSchema, 'ZodDefault')` obsolete. Possibly the one for zod effects too.

samchungy:
Resolves #60


src/openapi-generator.ts
```diff
@@ -938,23 +980,35 @@ export class OpenAPIGenerator {
...
     if (isZodType(schema, 'ZodOptional') || isZodType(schema, 'ZodNullable')) {
-      return this.unwrapChained(schema.unwrap());
+      return this.unwrapChained(schema.unwrap(), defaultValue);
     }
 
     if (isZodType(schema, 'ZodDefault')) {
-      return this.unwrapChained(schema._def.innerType);
+      return this.unwrapChained(
+        schema._def.innerType,
+        this.getDefaultValue(schema)
```
samchungy, line 997: Always use the latest default value. Otherwise keep passing it down.

src/openapi-generator.ts
```diff
@@ -842,9 +869,32 @@ export class OpenAPIGenerator {
...
+      return this.getDefaultSchema(zodSchema.unwrap());
+    }
+
+    if (isZodType(zodSchema, 'ZodEffects')) {
+      return this.getDefaultSchema(zodSchema._def.schema);
+    }
+
+    if (isZodType(zodSchema, 'ZodDefault')) {
+      return (
+        ('defaultValue' in zodSchema._def && zodSchema._def.defaultValue()) ||
```
georgyangelov, line 886: Is this not properly typed on the Zod size that we need the `in` check?

georgyangelov: Also I assume we want to use `?? undefined` because defaults of `0` would be ignored. We can add a test for that as well

src/openapi-generator.ts
```diff
@@ -653,9 +674,10 @@ export class OpenAPIGenerator {
     };
   }
 
-  private toOpenAPISchema(
-    zodSchema: ZodSchema<any>,
-    isNullable: boolean
+  private toOpenAPISchema<T>(
+    zodSchema: ZodSchema<T>,
+    isNullable: boolean,
+    defaultValue?: T
```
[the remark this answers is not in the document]
georgyangelov: Looks prettier :)

georgyangelov (changes requested):
Nice, thanks!

src/openapi-generator.ts
samchungy: Ah good point. it's because of the stupid && bit which could be false. I was trying to be clever. I'll fix that up

samchungy: Ah looks like when I refactored I no longer needed the `in`. Nice one

georgyangelov (approved):

georgyangelov:
Thank you for the change!

samchungy:
Heyo @georgyangelov @AGalabov could we cut a release some time soon? Not urgent but I would love to get some of the new schema features into my repos 😃

AGalabov:
@samchungy yup same. But having in mind it will be 3.0.0 we wanted to merge all our needs. We just need that last PR that is open.
We'll try to get it done by the end of today/week. By the way if you have any input there - please share it 😃

PR #66, opened 2022-11-16
Fix referencing schema with unrecognized type

blagowestas:


src/openapi-generator.ts
```diff
@@ -660,6 +657,21 @@ export class OpenAPIGenerator {
     };
   }
 
+  private getOpenAPISchema<T>(
```
AGalabov, line 660: I am kind of confused here - I like that we have it as a function. But the name sounds quite generic and reusable. Another thing is that we've already unwrapped the schema in `generateSimpleSchema` but we are doing it here as well. So I think there are 2 options:
1. Keep the function but make it something very specific and pass the appropriate arguments directly.
2. Keep the function more generic, reuse it in other places and try to reduce the duplication of code - for example the unwrap.

PR #70, opened 2022-11-21
Fix Route types to support strict objects

samchungy:
Currently the ZodObject<any> type defaults the second value to "strip" which means that "strict" types are not supported.

Substitute usage of `ZodObject<any>` with `AnyZodObject` type.

### DIFF
--- a/spec/routes.spec.ts
+++ b/spec/routes.spec.ts
@@ -302,6 +302,28 @@ const routeTests = ({
       ]);
     });
 
+    it('supports strict zod objects', () => {
+      const routeParameters = generateParamsForRoute({
+        request: {
+          query: z.strictObject({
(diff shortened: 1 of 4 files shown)


spec/routes.spec.ts
```diff
@@ -302,6 +302,28 @@ const routeTests = ({
       ]);
     });
 
+    it('supports strict zod objects', () => {
+      const routeParameters = generateParamsForRoute({
+        request: {
+          query: z.strictObject({
```
samchungy, line 308: Before this change this throws a type error.

AGalabov: Nice! Thank you!

AGalabov (approved):

samchungy:
Thanks for the release 🙌

PR #75, opened 2022-12-04
Fix exclusive minimum and maximum checks for numbers

JanTvrdik:


src/openapi-generator.ts
```diff
@@ -704,15 +704,45 @@ export class OpenAPIGenerator {
...
         ...this.mapNullableType(
           zodSchema.isInt ? 'integer' : 'number',
           isNullable
         ),
-        minimum: zodSchema.minValue ?? undefined,
-        maximum: zodSchema.maxValue ?? undefined,
         default: defaultValue,
       };
+
+      if (this.openApiVersionSatisfies(this.openAPIVersion, '3.1.0')) {
```
AGalabov, line 715: A couple of comments here:
1. Can you please extract this in a private function. Instead of the mutation approach you can just use the same approach as `mapNullableType` for example. Basically a method like `getNumberChecks` that would return a small object with `minimum`/`maximum`/`exclusiveMinimum`/`exclusiveMaximum`. And we can add the result into the same return above.
2. In my opinion the `this.openApiVersionSatisfies` check should be handled (and duplicated) within the separate cases instead of duplicating the for loop. The idea is that precisely min and max checks are affected by the OpenApi version. The integer check for example is not. And same can apply for future checks added to `zod` as well. So we should rather always be iterating over the array of checks but only apply the version check where needed (currently both places)

spec/types/number.spec.ts
```diff
@@ -19,4 +19,68 @@ describe('number', () => {
       Literal: { type: 'number', enum: [42] },
     });
   });
+
+  it('supports minimum in open api 3.0.0', () => {
```
AGalabov, line 23: Nice!

AGalabov (commented):
@JanTvrdik  thank you for the PR. Left a cosmetic comment but otherwise looks solid.

src/openapi-generator.ts
JanTvrdik: rewritten based on your feedback

AGalabov:
@JanTvrdik thanks once again. Just created a release 3.3.0 since I actually thought of this as more of a "feature" than a fix.

PR #78, opened 2022-12-12
Add ZodString Datetime support

samchungy:
Heya 🚤 

Since I added this bad boy to Zod, I figured I should be the one to implement it here. Not sure how we want to fix the type support here.

We could extend Zod in this library or we could just bump up the dev deps as this code should work with both versions. Happy with either, just let me know 😄

### DIFF
--- a/src/openapi-generator.ts
+++ b/src/openapi-generator.ts
@@ -587,7 +587,7 @@ export class OpenAPIGenerator {
       return 'uri';
     }
 
-    if (zodString.isDatetime()) {
+    if (zodString?.isDatetime()) {
       return 'datetime';
     }
 
--- a/src/openapi-generator.ts
+++ b/src/openapi-generator.ts
@@ -588,7 +588,7 @@ export class OpenAPIGenerator {
     }
 
     if (zodString.isDatetime?.()) {
-      return 'datetime';
+      return 'date-time';
     }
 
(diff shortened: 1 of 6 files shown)


src/openapi-generator.ts
```diff
@@ -587,6 +587,10 @@ export class OpenAPIGenerator {
       return 'uri';
     }
 
+    if (zodString.isDatetime?.()) {
```
samchungy, line 590: Optional chaining will mean backwards compatibility

AGalabov:
@samchungy sorry ... I somehow missed that there is a PR open among all other issue discussions 🤦. I'll have a look in a bit

src/openapi-generator.ts
```diff
@@ -587,6 +587,10 @@ export class OpenAPIGenerator {
       return 'uri';
     }
 
+    if (zodString.isDatetime) {
```
AGalabov, line 590: I love when I see that it's as easy as that to add a new feature :bow:

AGalabov (approved):

AGalabov:
@samchungy I am okay with keeping the peer dependencies untouched and just bumping the dev dependencies for now. I'll be merging the PR and scheduling a release in a bit (once again :smile:). Thank you!

PR #92, opened 2023-02-13
For 'nativeEnum', should be 'type: integer', not 'type: number

wiz-kakiuchi:
For example, if I use 'openapi-generator for php' to automatically generate a .php file from 'openapi-docs.yml', the 'nativeEnum' part is converted to PHP's float type.
Normally, numeric Enums should be of type int, not float.


AGalabov (approved):
Sounds perfectly valid. Thank you for the PR

AGalabov:
@wiz-kakiuchi just published a 4.2.0 version with your change

PR #93, opened 2023-02-20
Add discriminator mapping for objects using enums as discriminator

samchungy:
Currently throws an error so this should fix that :)

ps - downloads for this have exploded wew 🚀


georgyangelov (approved):
Nice, thanks!

AGalabov (approved):
Looks good! I'll merge it and tag it asap. Also the downloads have really skyrocketed ... not sure how/why or if it is actual traffic, but we take it :smile:

Issue #99, opened 2023-03-01
API improvement suggestion


EmilIvanov:
Hi, thank you for the great library. I'm playing around with it and I think it would be useful to be able to specify the schema name in the meta, e.g.:
```
const UserSchema = z.object({
  id: z.string().openapi({ example: '1212121', schemaName: 'ID' }),
  name: z.string().openapi({ example: 'John Doe', schemaName: 'Name' }),
  age: z.number().openapi({ example: 42, schemaName: 'Age' }),
}).openapi({ schemaName: 'User' })
```
This way schema registration can be done separately.

AGalabov:
Hi @EmilIvanov  thank you for the kind words. I am not quite sure what you'd want to achieve by doing that :thinking:? Is this meant to be an alternative to the current `registry.register('name', schema)` - where `name` would be the reference name under `component/schemas`?

If that is the case then how would you go about passing all schemas into the `OpenAPIGenerator`? And if it not that, could you please elaborate some more on your proposal and the added benefits?

EmilIvanov:
Thank you for the quick response. Here's the `tl;dr;`

The OpenApi definitions only change when your API changes. This means they can be generated just once at build time and then served statically. Imagine that you have `routes.ts` that exports your route definitions, `server.ts` that will serve them and a  `generate-openapi.ts` script that will import the routes and produce a `swagger.json`. I've implemented (a way worse version of) this using [joi](https://www.npmjs.com/package/joi) and [joi-to-swagger](https://www.npmjs.com/package/joi-to-swagger) which uses the schema's meta. The idea is that you just pass the route definitions to the generator function. It can register any schema on the fly the first time it encounters it and then just ref if. You don't really need to register any schema that's not on a path as it's not really a part of your API (though I'm not arguing that it's useful to be able to produce definitions for schemas that are not on your paths, it's just that they should not be mandatory at path definition generation).

And so, if I can put the `refId` in the meta I can use that to register it in that `generate-openapi.ts` script and omit this work in my application on startup. I'm fine with still having to register all the schemas - I can just import them and loop over them before I do the generation. Ideally, the library API (and that's just an opinion, really) would look like

```js
const { paths, components } = generatePaths({ paths, components });
```

As an added bonus, you can have a schema just lying in a file. If you need to register it at definition time, then you need to have a registry instance available and so now you need to export a function of the registry to a schema.

Hopefully at least some of this makes sense.

samchungy:
[...]

PR #101, opened 2023-03-04
Refactor OpenAPI Generator

samchungy:
Was trying to unwrap some of the code, looks like we have some duplicate code which doesn't do anything.

This will help with https://github.com/asteasolutions/zod-to-openapi/issues/99#issue-1604769637

### DIFF
--- a/src/openapi-generator.ts
+++ b/src/openapi-generator.ts
@@ -333,7 +333,7 @@ export class OpenAPIGenerator {
     const required =
       !this.isOptionalSchema(zodSchema) && !zodSchema.isNullable();
 
-    const schema = this.generateSimpleSchema(zodSchema);
+    const schema = this.generateSchemaDefinition(zodSchema);
 
     return {
       in: paramLocation,
@@ -347,7 +347,7 @@ export class OpenAPIGenerator {
   /**
    * Generates an OpenAPI SchemaObject or a ReferenceObject with all the provided metadata applied
    */
-  private generateSimpleSchema<T>(
+  private generateSchemaDefinition<T>(
     zodSchema: ZodSchema<T>
   ): SchemaObject | ReferenceObject {
     const innerSchema = this.unwrapChained(zodSchema);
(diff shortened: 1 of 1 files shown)


src/openapi-generator.ts
```diff
@@ -333,7 +333,7 @@ export class OpenAPIGenerator {
     const required =
       !this.isOptionalSchema(zodSchema) && !zodSchema.isNullable();
 
-    const schema = this.generateSimpleSchema(zodSchema);
+    const schema = this.generateSchemaDefinition(zodSchema);
```
AGalabov, line 336: I don't think I can agree with this one. Right now this change would enforce a schema being defined into `schemaRefs` and would then be saved in `paramRefs` (in `generateParameterDefinition`). That was the overall idea behind having a `generateSimpleSchema` that does not bother "saving the result ref".

src/openapi-generator.ts
```diff
@@ -1113,12 +1098,12 @@ export class OpenAPIGenerator {
 
   private applySchemaMetadata(
     initialData: SchemaObject | ParameterObject | ReferenceObject,
-    metadata: Partial<ZodOpenAPIMetadata>
+    metadata?: Partial<ZodOpenAPIMetadata>
```
AGalabov, line 1101: I am not a big fan of this even if it helps with some of the code.

Basically right now we have a method saying "apply metadata" that can work without metadata - that's very odd in my opinion. I guess we can work around this by renaming the method :thinking:. Not sure what could be a suitable name.

And one more thing - if we do end up renaming it and sticking with the optional, there are a couple more instances where a ternary operator was used to cover for undefined metadata.

AGalabov (commented):
@samchungy I get the idea and deep down I know that there is some duplication so I am very much up for a refactoring. However, I did raise some questions. And additionally can you share how you envision this helping with #99?

samchungy:
> @samchungy I get the idea and deep down I know that there is some duplication so I am very much up for a refactoring. However, I did raise some questions. And additionally can you share how you envision this helping with #99?

How do I register a schema without calling `.register`? I'm thinking that, while we traverse the Zod object - whenever we run into `refId` or `schemaRef` or TBD, we add it to `this.schemaRefs` and create a ref to it. In my head I think that makes sense

src/openapi-generator.ts
samchungy: Oh I see I see, let me rethink this :)

PR #107, opened 2023-03-16
Feature/#103 add support for tuples

AGalabov:
* Added implementation of `ObjectSet` and a `uniq` function that utilizes it
* Added support for tuples through `array`

### DIFF
--- a/src/lib/object-set.ts
+++ b/src/lib/object-set.ts
@@ -34,11 +34,13 @@ export class ObjectSet<V> {
   private buckets = new Map<number, V[]>();
   put(value: V) {
     const hashCode = this.hashCodeOf(value);
+
     const itemsByCode = this.buckets.get(hashCode);
     if (!itemsByCode) {
       this.buckets.set(hashCode, [value]);
       return;
     }
+
     const alreadyHasItem = itemsByCode.some(_ => isEqual(_, value));
     if (!alreadyHasItem) {
       itemsByCode.push(value);
@@ -47,6 +49,7 @@ export class ObjectSet<V> {
 
   contains(value: V): boolean {
     const hashCode = this.hashCodeOf(value);
+
     const itemsByCode = this.buckets.get(hashCode);
     if (!itemsByCode) {
       return false;
@@ -62,6 +65,7 @@ export class ObjectSet<V> {
     let totalBuckets = 0;
     let totalValues = 0;
     let collisions = 0;
+
     for (const bucket of this.buckets.values()) {
       totalBuckets += 1;
       totalValues += bucket.length;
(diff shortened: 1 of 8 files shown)


AGalabov:
An interesting thing to consider is the [following comment](https://stackoverflow.com/a/57465199). It suggests that for OpenApi 3.1.0 we can use `prefixItems`. However the library `openapi3-ts` does not support it and neither does `openapi31-ts`. I am yet to debunk whether that is a valid OpenApi syntax. If it is I'd provide the needed type definitions to allow it manually (possibly open PRs to the respected libraries as well). Let me know your opinion on the matter

src/lib/object-set.ts
```diff
@@ -0,0 +1,95 @@
...
+      totalBuckets += 1;
+      totalValues += bucket.length;
+      if (bucket.length > 1) {
+        collisions += 1;
+      }
+    }
+    const hashEffectiveness = totalBuckets / totalValues;
+    return { totalBuckets, collisions, totalValues, hashEffectiveness };
+  }
+  private hashCodeOf(object: any): number {
```
georgyangelov, line 75: I'd put some blank lines here and there :D

src/openapi-generator.ts
```diff
@@ -853,6 +854,35 @@ export class OpenAPIGenerator {
...
+          type: 'array',
+          items: uniqueSchemas[0],
+          minItems: tupleLength,
+          maxItems: tupleLength,
+        };
+      }
+
+      return {
+        ...this.mapNullableType('array', isNullable),
+        // TODO: Compact is extra
```
georgyangelov, line 877: Sorry, what?

georgyangelov (approved):

src/openapi-generator.ts
AGalabov: There was a `compact` which was extra :smile: Now there is no compact ... but there was not TODO for the comment being extra. I'll remove it, thanks

samchungy:
Btw looks like there's a better way to represent it in 3.1?

https://stackoverflow.com/questions/57464633/how-to-define-a-json-array-with-concrete-item-definition-for-every-index-i-e-a#57465199

AGalabov:
@samchungy I've mentioned that exact Stack Overflow comment in my [initial comment](https://github.com/asteasolutions/zod-to-openapi/pull/107#issuecomment-1472159200). I didn't really manage to get a clarification that this is valid for OpenApi 3.1 - i.e with an online validator/documentation renderer. Please let me know if you find any.

samchungy:
> @samchungy I've mentioned that exact Stack Overflow comment in my [initial comment](https://github.com/asteasolutions/zod-to-openapi/pull/107#issuecomment-1472159200). I didn't really manage to get a clarification that this is valid for OpenApi 3.1 - i.e with an online validator/documentation renderer. Please let me know if you find any.

I'll play around with something tomorrow at work and get back to you if it works 😄

PR #111, opened 2023-03-20
use zod schema description when generating components instead of only .openapi

AGalabov:


src/openapi-generator.ts
```diff
@@ -1137,7 +1137,28 @@ export class OpenAPIGenerator {
...
+     * That description should be used when generating an OpenApi schema.
+     * The `??` bellow makes sure we can handle both:
+     * - schema.describe('Test').optional()
+     * - schema.optional().describe('Test')
+     */
+    const zodDescription = zodSchema.description ?? innerSchema.description;
+
+    /**
+     * A description provided from .openapi() should be taken with higher
+     * precedence. Note that there is an overload of .describe() that uses
```
georgyangelov, line 1151: Shouldn't we remove the overload? What purpose does it have now?

AGalabov: Well that is a good question. This is the better implementation in my opinion. In case someone uses the .describe code with `extendZodWithOpenApi` then this would also mean that you can do `.openapi({ description: 'Description' }).describe('Alternative')`. If there is no such implementation you cannot do it. Note: In my opinion it is also valid to do this:  `.describe('Description').openapi({ description: 'Alternative'})` => we cannot just say "zod describe is before openapi or the other way round".

Issue #115, opened 2023-04-14
Support for ZodPipeline


jesinmat:
Hello,
I'm using Zod Pipelines according to the [Zod documentation](https://zod.dev/?id=you-can-use-pipe-to-fix-common-issues-with-zcoerce) to safely handle dates. I have the following object:

```
export const WorkerSchema = z
  .object({
    name: z.string().min(1),
    available: z.array(z.date().or(z.string().min(1).pipe(z.coerce.date()))),
  })
  .strict();
```

This way, I can either pass date directly or a string of length at least 1 that can be converted to a date.
Using this with zod-to-openapi, the following error is generated:

```
UnknownZodTypeError {
  message: 'Unknown zod object type, please specify `type` and other OpenAPI props using `ZodSchema.openapi`.',
  data: {
    currentSchema: { in: [ZodString], out: [ZodDate], typeName: 'ZodPipeline' },
    schemaName: undefined
  }
}
```

Since both ZodString and ZodDate are specified, is it possible to convert this automatically?
What should I do to overcome this issue in the meantime?
And since I use this approach several times in my project, can I specify this conversion method in one place for all of them?

AGalabov:
Hi @jesinmat, thanks for using our library and pointing this out. `.pipe()` is relatively new so we've not looked into it before.

Overall I'd need to check if there is persistent way to make sure that the type is correctly inferred. For example `.transform` cannot be handled within `zod-to-openapi`. From what I see in the error I am optimistic.

As for a fix in the meantime - yes there is one. As the message of the error suggests you can just pass `.openapi({ type: 'string' ]})` to your schema. And for reusability I guess you can just use regular programming means and do something like:

```ts
// In some common file:
export const dateSchema = z.date().or(z.string().min(1).pipe(z.coerce.date()));

// In your current file:
export const WorkerSchema = z
  .object({
    name: z.string().min(1),
    available: z.array(dateSchema),
  })
  .strict();
```

Let me know if that works for you. And I'll see if we can provide an out of the box support for pipelines.

jesinmat:
Thank you for your quick reply and the pull request!
I ended up using `.openapi({ type: "string", format: "date" })` and it works well. This library has helped me a lot, glad to see it's continuing to improve.

AGalabov:
@jesinmat this should now be available in the [v4.6.0](https://github.com/asteasolutions/zod-to-openapi/releases/tag/v4.6.0) release

PR #118, opened 2023-04-15
#115 added support for ZodPipeline

AGalabov:
Based on the [documentation](https://zod.dev/?id=pipe) it seems that we can generate a documentation for `ZodPipeline`.

However we do have two options:
1. Use the `in` property saved within the zod schema (done in the PR). My thought process was that this is what an API would receive when this schema is used for validation.
2. Use the `out` property as it describes the target state (might not always be the case).

Overall I feel like `in` is the only correct option we can use. But saying that it does bring up the question why cannot we just use the same logic when handling the result of `.transform()`. @georgyangelov any thoughts would be appreciated.

### DIFF
--- a/spec/modifiers/pipe.spec.ts
+++ b/spec/modifiers/pipe.spec.ts
@@ -2,10 +2,6 @@ import { z } from 'zod';
 import { expectSchema, registerSchema } from '../lib/helpers';
 
 describe('pipe', () => {
-  z.string()
-    .transform(val => val.length)
-    .pipe(z.number().min(5));
-
   it('can generate schema for pipes', () => {
     expectSchema(
       [
--- a/spec/modifiers/pipe.spec.ts
+++ b/spec/modifiers/pipe.spec.ts
@@ -1,0 +1,39 @@
+import { z } from 'zod';
+import { expectSchema, registerSchema } from '../lib/helpers';
+
+describe('pipe', () => {
+  z.string()
+    .transform(val => val.length)
+    .pipe(z.number().min(5));
+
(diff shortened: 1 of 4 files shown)


src/openapi-generator.ts
```diff
@@ -974,6 +974,10 @@ export class OpenAPIGenerator {
       };
     }
 
+    if (isZodType(zodSchema, 'ZodPipeline')) {
+      return this.toOpenAPISchema(zodSchema._def.in, isNullable, defaultValue);
```
georgyangelov, line 978: Well this was easy 😄

georgyangelov (approved):
Looks good. As far as `transform` goes - I agree, we can use the input schema. I was under the impression that we couldn't extract the in schema in that case and that's why we weren't supporting it.

spec/modifiers/pipe.spec.ts
```diff
@@ -0,0 +1,39 @@
+import { z } from 'zod';
+import { expectSchema, registerSchema } from '../lib/helpers';
+
+describe('pipe', () => {
+  z.string()
+    .transform(val => val.length)
+    .pipe(z.number().min(5));
```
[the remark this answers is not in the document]
AGalabov: I got what the problem is :smile:

src/openapi-generator.ts
nidomiro: Are you sure that `in` is the right thing to use here?

I would expect something like this:
- Use `in` for `parameters` as this is the possible input, from which an instance can be created
- Use `out` for `schemas` as this is the resulting type that will come out

PR #119, opened 2023-04-17
#99 automatic registration

AGalabov:
This is a fully working RFC that implements the ideas of #99. However this PR also includes a non trivial amount of refactoring - most of which was needed in order to simplify the development process - hence why it stepped on a foundation made in [this PR](https://github.com/asteasolutions/zod-to-openapi/tree/bugfix/%23104-nested-description-not-being-used).

Side Note: I highly recommend reviewing commit by commit

## There are a couple of things that can/must be answered before this PR is merged (apart from code review):

1. :heavy_check_mark: Initially I created a way to execute all tests with both "registration approaches". That helped me figure out some minor discrepancies. However at this moment in time the two are pretty much the same, since the registry uses `.openapi(refId)` internally. This means that the `registrationTypeDescribe` and all related changes might be obsolete and the tests entirely duplicated. Personally I am all for deleting them
2. :heavy_check_mark: Decide if the overload of `.openapi` approach is the best suited option. It was the most natural option I could think of.
3. :heavy_check_mark: Should we keep providing a deprecated `internal_openapi`? It is currently unnecessary for the inner working of the library and it is not supposed to be used outside of it anyways. This is a breaking change. However I was always planning this to be a major version anyways.
4. :heavy_check_mark: Right now parameter schemas are being automatically registered into `components/schemas`. By OpenApi standards this is fine and I also think that this is the correct behavior. A parameter object would still be added into `components/parameters` if `registry.registerParameter` is used. However its `schema` key would just reference the automatically generated schema from  `components/schemas`.
5. Right now `ZodEffect` schemas would not generate their internal schema. (see the skiped tests in `auto-register.spec.ts`)
6. :heavy_check_mark: Address all other TODO comments in the code 
7. :heavy_check_mark: When the design decisions and limitations are agreed upon the README.md file should be updated.


georgyangelov:
1. I really like the way `<schema>.openapi(<name>)` looks!
2. Do we want to have support for putting the components in a different "folder" instead of `components/schemas`? We can probably do that easily by allowing `<name>` to contain slashes or start with `#`, e.g. `<schema>.openapi('#/components/somethingelse/SchemaName')`?
3. I'm not a fan of the changes in tests, I think we can think about what the registry gives us in addition to the `.openapi(<name>)`. Probably it gives us the ability to register schemas without using them in routes? But in that case we can simplify the registry somehow? Will have to think about it a bit more. And in principal having two ways to do something may not be a good practice since we'll have to support both, but maybe that's ok.
4. Why do we need the auto-register spec? Can we have that be part of the other test files?
5. We can remove the `internal_openapi` method but I vote to keep the `refId` under `_internal` as you've done now so that any extra openapi extension keys do not override this.
6. I'll review the openapi-generator file tomorrow if that's ok with you

AGalabov:
@georgyangelov thank you for the initial review.
1. The folder naming idea is interesting. I am open to consider the possibilities. However I think we would be better off handling it as a separate PR :thinking: for brevity. Since it is a "new feature" of sorts it can even go as a release of its own if needed.
2. As I said, I think it would be best to simplify the tests as well. And right now I wouldn't say that they are different methods actually since the registry uses the `.openapi` internally. And you can still pass a random unused schema into the generator (no need for a route). The main thing that the registry gives in my opinion is a common way to "collect schemas, routes, etc" so that you can simply pass them. The same can be achieved with a random array that we push into. However the other thing that it does is that it allows for custom components with TS and parameter registration (something that you cannot do in any other way right now)
3. I don't necessarily insist on `auto-register.spec` being kept that way. My main thought about it was that it is easier for me to develop and would also be much easier for anyone reviewing (since I knew there would quite a few) to take a look at the API and the added features.

Looking forward to other comments.

EmilIvanov:
@AGalabov I pulled this changed locally and was able to "unspagettify" my OpenAPI generation script, and greatly reduce is to this trivial script:

```typescript
export const getDocumentGenerator = (version: OpenApiVersion, routes: RouteDefinition[]) => {
  const registry = new OpenAPIRegistry();

  registry.registerComponent('securitySchemes', 'bearerAuth', {
    type: 'http',
    scheme: 'bearer',
    bearerFormat: 'JWT',
  });

  for (const route of routes) {
    registerRoute(registry, route);
  }

  return new OpenAPIGenerator(registry.definitions, version);
};
```

I have some utility specific to my project in `registerRoute` but it's essentially just a call to `registry.registerPath(route)`.

It seems to be what I was looking for, so thank you!

One final, sort-of-unrelated request (can be implemented separately). Can you export `OpenApiVersion`, cause right now I have to import it from `@asteasolutions/zod-to-openapi/dist/openapi-generator`?

Thanks,

src/openapi-generator.ts
```diff
@@ -346,87 +363,105 @@ export class OpenAPIGenerator {
...
+
+    return metadata?.metadata
+      ? this.applySchemaMetadata(result, metadata.metadata)
+      : omitBy(result, isNil);
+  }
+
   /**
    * Generates an OpenAPI SchemaObject or a ReferenceObject with all the provided metadata applied
    */
   private generateSimpleSchema<T>(
```
georgyangelov, line 383: Idea for a refactoring at some point (not now) - currently I believe we do a lot of duplicated unwrapping, for example we call `unwrapChained` at multiple points on the same schema. It may be possible to get some good performance gains by reorganizing so that these lookups are cached or reused in some other way.

georgyangelov (approved):

georgyangelov:
Looks good, I like how it turned out!

AGalabov:
@EmilIvanov thank you for the input. That does seem quite neet so it is definitely a sign that we've done something right. I'll take a look at the code one final time + catch up with the other PRs and write up a new README file and hopefully we should be able to merge and release this in the upcoming week. Ideally I'd want to pair it up as a single release with #125

Issue #122, opened 2023-04-25
Upgrade openapi3-ts to latest to fix vulnerability in YAML


venkateshpai92:
There is a Vulnerability [CVE-2023-2251](https://nvd.nist.gov/vuln/detail/CVE-2023-2251) for the npm package [YAML](https://www.npmjs.com/package/yaml)

`openapi3-ts` latest version (4.1.2) fixes this vulnerability as it uses the YAML version `2.2.2`.

Could you please upgrade `openapi3-ts` to `4.1.2`?

AGalabov:
@venkateshpai92  it should be fixed in the latest release - [v4.6.2](https://github.com/asteasolutions/zod-to-openapi/releases/tag/v4.6.2)

PR #125, opened 2023-04-26
Fix vulnerability in yaml

AGalabov:
Note: This update lead to changes in the types of `openapi3-ts` related to the addition of OpenApi 3.1. I've currently take the "easier" approach for this update - i.e casting in 1 place in the generator and in a couple more in the tests.

In my opinion a better approach would be to do something like:
1. Separate implementation/type overloads of every method and schema based on the provided `openAPIVersion`.
2. Decision on what needs to be exposed as option for `.openapi`
  - It could be a "union" of schemas - that would however introduce new behavior to be implemented.
  - Or we can keep the current "old" schema (which "forbids" certain usages that are valid for 3.1)
  - We could also expose two overloads - something like `.openapi` and `.openapi31`. Again that would introduce new properties and cases that we should handle in the code

@georgyangelov @samchungy I am once again asking for opinion

Fixes: #122

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "4.6.0",
       "license": "MIT",
       "dependencies": {
-        "openapi3-ts": "^3.1.1"
+        "openapi3-ts": "^4.1.2"
       },
       "devDependencies": {
         "@types/jest": "^29.2.5",
@@ -17,7 +17,7 @@
         "prettier": "^2.7.1",
         "ts-jest": "^29.0.3",
         "typescript": "^4.6.3",
-        "yaml": "^2.0.0",
+        "yaml": "^2.2.2",
         "zod": "^3.20.2"
       },
       "peerDependencies": {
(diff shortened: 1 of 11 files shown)


samchungy:
My 2c is if you want a quick way an intersection of the two types should be sufficient for now?

georgyangelov:
I feel like with the next version since we'll be changing the API anyway for auto-registration, we can also change the extension mechanism a bit. Make it more like this:

```ts
import { z as zod } from 'zod' 
import { extendZod } from '@asteasolutions/zod-to-openapi'

export const z = extendZod(zod, '3.1')
```

and we can provide two overloads for `extendZod` which produce the 3.0 or the 3.1 version of `.openapi`.

What do you think? Actually with that we can make a separate library `@asteasolutions/zod-to-openapi-extension` which we can depend on and that can include just `extendZod` and `.openapi`, so that the generator can be development-only.

AGalabov:
@georgyangelov this sounded great at first. However that does not solve our problem. The implementation of `.openapi` is actually the same among the two versions (so there is no need of an overload there). The thing that needs to be changed is the `declare module 'zod'` definition so that we can have different typescript support for it.

PR #127, opened 2023-04-28
Separate openapi versions into classes

AGalabov:
This is how I imagine the code should look like. The problem I am currently facing is having two separate zod extensions - that allow to do something like:
```ts
import { extendZodWithOpenApi } from '@asteasolutions/zod-to-openapi/v3.1';

z.string().openapi(/* The type here should be those of SchemaObject from v3.1 */)
```

and 

```ts
import { extendZodWithOpenApi } from '@asteasolutions/zod-to-openapi/v3.0;

z.string().openapi(/* The type here should be those of SchemaObject from v3.0 */)
```
I cannot get this working without breaking multiple places in the code (for "incompatibility"). To be honest - the incompatibility is correct because of the discrepancies between the two versions. The most obvious solution to the problem would be to have two completely separate classes with the shared code being in an "unrelated" class or separate functions that work with generic types. However it seems to me that this would still enforce a lot of code duplication.


src/openapi-generator.ts
```diff
@@ -60,21 +60,27 @@ const openApiVersions = ['3.0.0', '3.0.1', '3.0.2', '3.0.3', '3.1.0'] as const;
...
-  OpenAPIObject,
-  'paths' | 'components' | 'webhooks' | 'openapi'
->;
-
 interface ParameterData {
   in?: ParameterLocation;
   name?: string;
 }
 
+// TODO: Can I improve upon the anys
```
georgyangelov, line 68: Well, at least the return types are actually "any" - can be anything. Or map them to `unknown`

src/openapi-generator.ts
```diff
@@ -719,44 +705,19 @@ export class OpenAPIGenerator {
...
     isNullable: boolean
   ): (SchemaObject | ReferenceObject)[] {
-    if (isNullable) {
-      if (this.openApiVersionSatisfies(this.openAPIVersion, '3.1.0')) {
-        return [...objects, { type: 'null' }];
-      }
-      return [...objects, { nullable: true }];
-    }
-    return objects;
+    isNullable;
```
georgyangelov, line 712: hm?

src/v3.1/openapi-generator.ts
```diff
@@ -0,0 +1,71 @@
...
+    private definitions: (OpenAPIDefinitions | ZodSchema)[],
+    private openAPIVersion: OpenApiVersion
+  ) {
+    const specifics = new OpenApiGeneratorV31Specifics();
+    this.generator = new OpenAPIGenerator(this.definitions, specifics);
+  }
+
+  generateDocument(config: OpenAPIObjectConfigV31): OpenAPIObject {
+    // TODO: Fix the casts here. Potentially create "reusable" types based on openapi3-ts
+    const baseDocument = this.generator.generateDocumentData() as Pick<
```
georgyangelov, line 37: What is the error here? What cannot be casted?

georgyangelov (approved):

PR #128, opened 2023-04-28
added support for zod object for response headers

### LINKED ISSUE #116: ZodObjects not accepted in some places
Hey, congrats for the library. We use it in a formative open source project, but there is not a lot of documentation and we can't find how to achieve some things.

In particular, we would expect to be able to pass ZodObjects to header schemas and to the 'oneOf' key in responses to specify more than one response per code, but they both won't accept them and we don't know what else to try. 

[This is the file](https://github.com/IT-Academy-BCN/ita-wiki/blob/8fdfa9c9688d2783e387415d30bd68458dad5826/back/src/openapi/routes/authRouter.ts) that we are working on with the issues commented out, any help would be appreciated.

AGalabov:
Resolves #116.

Side note: right now there are 3 approaches to generate response headers:
1. Directly specifying a `HeadersObject` through the response 
```
responses: {
  204: {
    description: 'Success',
    headers: { 'Set-Cookie': { schema: { type: 'string' } } },
  },
},
```
2. Using custom components:
```
const apiKeyHeader = registry.registerComponent('headers', 'api-key', {
      example: '1234',
      required: true,
    });

// ...
responses: {
  200: { description: 'Sample response', headers: { 'x-api-key': apiKeyHeader.ref }  }
};
```
3. Using ZodObject:
```
responses: {
  204: {
    description: 'Success',
    headers: z.object({
      'Set-Cookie': z.string().openapi({ example: 'token=test' }),
    }),
  },
},
```

However I do feel that all are valid because:
1. Obviouslly we should allow plain documentation to be written as everywhere else in the codde
2. The custom schemas are a mechanism that can be utilized for response headers but are not necessarily made for that
3. Adding support for ZodObject seems logical since this is something that could be validated and it is a convenience to support.

### DIFF
--- a/spec/routes/index.spec.ts
+++ b/spec/routes/index.spec.ts
@@ -1,8 +1,7 @@
-import { z, ZodSchema } from 'zod';
-import { OperationObject, PathItemObject } from 'openapi3-ts';
+import { z } from 'zod';
 import { OpenAPIGenerator } from '../../src/openapi-generator';
-import { OpenAPIRegistry, RouteConfig } from '../../src/openapi-registry';
-import { createTestRoute, registerSchema, testDocConfig } from '../lib/helpers';
+import { OpenAPIRegistry } from '../../src/openapi-registry';
+import { createTestRoute, testDocConfig } from '../lib/helpers';
 
 const routeTests = ({
   registerFunction,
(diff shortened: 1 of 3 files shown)


georgyangelov (approved):
Nice

Issue #132, opened 2023-05-09
OpenAPI metadata getting lost when specified inside z.deepPartial() object schema


fomin-alexander:
## Use Case
We're trying to define `deepPartial()` schema for our query parameters (all parameters are optional):

```ts
{
  request: {
    query: z
      .object({
        filter: z.object({ test: z.string() }).openapi({ param: { style: 'deepObject' } }),
      })
      .deepPartial(),
  },
}
```

## Expected Result
```js
{
  "schema": {"type": "object", "properties": {"test": {"type": "string"}}},
  "required": false,
  "style": "deepObject", // <-- Expected to see metadata added by openapi()
  "name": "filter",
  "in": "query"
}
```

## Actual Result
```js
{
  "schema": {"type": "object", "properties": {"test": {"type": "string"}}},
  "required": false,  // <-- deepPartial() applied, but somehow removed metadata
  "name": "filter",
  "in": "query"
}
```

## Workaround
Instead of `deepPartial()` we use `optional()` schema for every nested attribute.
```ts
{
  request: {
    query: z.object({
      filter: z
        .object({ test: z.string().optional() })
        .optional()
        .openapi({ param: { style: 'deepObject' } }),
    }),
  }
}
```
It's ok, but it seems like valid Zod schema for query didn't worked.

AGalabov:
@fomin-alexander thank you for the PR. Although your PR is fixing the issue, I felt like it was not fixing it at the root cause - the overall problem was that we did not create a deep partial overload (in the `zod-extension` file). In general using `.openapi` is made to stack up/override metadata. So instead of "copying" it in the generator the correct fix was to preserve them when `deepPartial` is called. You can check [this PR:](https://github.com/asteasolutions/zod-to-openapi/pull/144)

fomin-alexander:
@AGalabov Thank you so much for paying attention to the issue and explaining your fix! Will check it soon!

PR #133, opened 2023-05-09
Bugfix/#132: Keep inner param metadata to fix deepPartial nested object params

fomin-alexander:
This PR supposed to fix [issue](https://github.com/asteasolutions/zod-to-openapi/issues/132) when openapi metadata is being list when using inside deepPartial object schema. The problem occurred because of calling `openapi()` on the top level schema:
```ts
        return this.generateParameter(
          schema.openapi({ param: { name: key, in: location } })
        );
```

The above call added metadata to the top level schema which could be `ZodOptional` in case of `deepPartial()`, so after that, when `getMetadata()` was called from `generateParameter()`, it didn't return metadata from inner object:
```ts
    const metadata = zodSchema._def.openapi
      ? zodSchema._def.openapi // <-- Metadata already exists at the top level when generating param
      : innerSchema._def.openapi;
```

To fix this, I added inner param metadata to the top level. Please check my solution and let me know if it works. Thanks!

### DIFF
--- a/spec/routes/parameters.spec.ts
+++ b/spec/routes/parameters.spec.ts
@@ -21,6 +21,37 @@ describe('parameters', () => {
       ]);
     });
 
+    it('generates a deepPartial object query parameter for route', () => {
+      const { parameters } = generateDataForRoute({
+        request: {
+          query: z
+            .object({
+              filter: z
+                .object({ test: z.string() })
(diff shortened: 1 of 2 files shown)


AGalabov:
A fix for the root cause was introduced in [this PR](https://github.com/asteasolutions/zod-to-openapi/pull/144)

PR #163, opened 2023-07-24
Use oneOf instead of anyOf for ZodUnion

### LINKED ISSUE #162: Use oneOf instead of anyOf for ZodUnion
I'm not an expert in OpenApi or Zod, but I have one good reason to use `oneOf` instead of `anyOf` - openapi-generator generates correct typescript-axios client code when `oneOf` is used and it generates wrong code when `anyOf` is used.

See https://github.com/OpenAPITools/openapi-generator/issues/12432#issuecomment-1647864275

In the linked comment I provided examples of OpenAPI specification generated by `zod-to-openapi`.

1.
```yaml
Component:
  anyOf:
    - $ref: "#/components/schemas/Component1"
    - $ref: "#/components/schemas/Component2"
```
generates empty interface:
```ts
export interface Component {
}
```

2.
```yaml
Component:
  type: object
  additionalProperties:
    anyOf:
      - $ref: "#/components/schemas/Component1"
      - $ref: "#/components/schemas/Component2"
```

combines all properties to single interface as described in original issue.

I updated specification file by changing all `anyOf:` to `oneOf` manually and after running openapi-generator-cli I got correct exports:
```ts
export type Component = Component1 | Component2;
```

Naktibalda:
Closes #162

### DIFF
--- a/spec/metadata-overrides.spec.ts
+++ b/spec/metadata-overrides.spec.ts
@@ -183,7 +183,7 @@ describe('metadata overrides', () => {
         type: 'object',
       },
       UnionTest: {
-        anyOf: [{ type: 'string' }, { $ref: '#/components/schemas/Empty' }],
+        oneOf: [{ type: 'string' }, { $ref: '#/components/schemas/Empty' }],
         description: 'Union with empty object',
       },
     });
--- a/spec/modifiers/catchall.spec.ts
+++ b/spec/modifiers/catchall.spec.ts
@@ -66,7 +66,7 @@ describe('catchall', () => {
               bonus: { type: 'number' },
             },
             additionalProperties: {
-              anyOf: [
+              oneOf: [
                 { type: 'boolean' },
                 { type: 'number' },
                 { type: 'string' },
--- a/spec/modifiers/pipe.spec.ts
+++ b/spec/modifiers/pipe.spec.ts
@@ -12,7 +12,7 @@ describe('pipe', () => {
       ],
       {
         PipedDate: {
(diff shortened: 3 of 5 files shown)


georgyangelov:
@Naktibalda Thank you very much for taking the time to open a PR! However, I don't think `oneOf` is correct in all cases. Note that the difference between `oneOf` and `anyOf` is that `oneOf` must match exactly 1 of the schemas, while `anyOf` can match multiple.

Consider the case of the following definition:

```ts
const A = z.object({ one: z.string(), two: z.string().optional() })
const B = z.object({ one: z.string() })

const Schema = z.union([A, B])
```

Then the input object `{ one: 'test' }` matches both schemas, so `anyOf` is correct while `oneOf` is not since it rejects the perfectly valid `{ one: 'test' }`.

I don't think we should be changing the OpenAPI schema we generate here since your original issue looks like a bug in the other library that reads the `anyOf`.

georgyangelov:
Yes, see https://swagger.io/docs/specification/data-models/oneof-anyof-allof-not/#difference

Naktibalda:
In my case it would be nice if Zod could produce an error if input matches multiple schemas, because it means that user provided too many parameters.

As a workaround I replaced `anyOf` with `oneOf` in generated yaml text:
```ts
const document = generator.generateDocument(config);
const specificationYaml = yaml.stringify(document);
const patchedSpecificationYaml = specificationYaml.replaceAll(
    /\banyOf:/g,
    "oneOf:"
  );
```

PR #165, opened 2023-07-25
feat: combine config.parameters with generated from config.request

### LINKED ISSUE #164: Combine config.parameters with ones generated from config.request
Hey 👋🏻. Currently when using `registerPath` method if both fields `RouteConfig.parameters` and `RouteConfig.request` are defined, `parameters` field is overwritten by the `request` field. 

Would it be possible to combine these two fields? 

I encountered this when i tried to do this 👇🏻 as cookies field is not currently present in the request object:

```
registry.registerPath({
      ...
      parameters: [
        {
          in: 'cookie',
          name: 'sessionCookie',
        },
      ],
      request: {
        query: z.object({
          id: z.string(),
        }),
      },
    });
```

which results in:
```
parameters: [
      {
        schema: { type: 'string' },
        required: true,
        name: 'id',
        in: 'query'
      }
]
```

instead of:

```
parameters: [
      { in: 'cookie', name: 'sessionCookie' },
      {
        schema: { type: 'string' },
        required: true,
        name: 'id',
        in: 'query'
      }
]
```

vosmol:
resolves #164

### DIFF
--- a/spec/routes/parameters.spec.ts
+++ b/spec/routes/parameters.spec.ts
@@ -294,6 +294,36 @@ describe('parameters', () => {
     });
   });
 
+  it('combines parameter definitions', () => {
+    const { parameters } = generateDataForRoute({
+      request: {
+        query: z.object({ request_queryId: z.string() }),
+        params: z.object({ request_paramsId: z.string() }),
+      },
+      parameters: [
+        { in: 'query', name: 'params_queryId' },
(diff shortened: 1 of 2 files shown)


AGalabov (approved):
@vosmol  thank you for opening up a PR. And sorry it took so long. Looks good! I'll merge it and release a new version for you

PR #180, opened 2023-10-21
WIP: force one of flag

fmmoret:
Hey, I am working with an API that I cannot add discriminators to, but downstream generators can't consume AnyOf properly in some cases. 

Here are some similar threads:
https://github.com/asteasolutions/zod-to-openapi/issues/162
https://github.com/asteasolutions/zod-to-openapi/pull/163

This PR is a proposal for an interface to let a user opt into a potentially unsafe representation of their schema.

If the direction is approved, I can add testing / ensure correctness of my implementation -- I have not manually checked it yet for correctness, but want feedback on the interface & approach.

### DIFF
--- a/src/openapi-generator.ts
+++ b/src/openapi-generator.ts
@@ -960,6 +960,14 @@ export class OpenAPIGenerator {
         return this.generateSchemaWithRef(optionToGenerate);
       });
 
+      const internalOptions = this.getOptions(zodSchema);
+      if (internalOptions.forceOneOf) {
+        return {
+          oneOf: this.mapNullableOfArray(schemas, isNullable),
+          default: defaultValue,
+        };
+      }
+
       return {
         anyOf: this.mapNullableOfArray(schemas, isNullable),
         default: defaultValue,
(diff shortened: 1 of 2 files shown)


AGalabov:
@fmmoret Thank you for opening a PR and sorry for taking long on this. I think I do agree that we can add such an option (as opposed to the previous implementation in #163).

A couple of pointers:
1. I think it makes sense to be consistent across the document => I would rather move the configuration at constructor level instead of through the `.openapi` method. I simply cannot envision why one would want to have two `ZodUnion` schemas with different signature within a single document.
2. We tend to add tests for all features/bugs that we handle. Please add some for this.

I'd also cc: @georgyangelov to share his opinion on the matter.

georgyangelov:
@fmmoret thank you for the contribution. 

@AGalabov I'm a bit on the fence for constructor parameter vs `.openapi`. On one hand the `.openapi` approach gives more control - you can use `oneOf` for some of the types (I also can't think of a reason to want that but that doesn't mean there isn't any). On the other hand - that way would mean going through all of your types and adding that to every single one where a union may be declared. The latter is I believe more of an issue so I support the global switch in the geenrator object approach - that would mean users won't forget to add the flag when the schemas change.

steven-t-h:
@fmmoret are you still pursuing this? I found myself also looking for this solution. I agree with your approach and I'm happy to jump in on this if needed.

PR #183, opened 2023-11-06
Added date-format to ZodDate

shaikegross:
I want to have my model be of date type but be rendered as date in openapi-generators. The only way to do this without having to hack the generator is to c;hange the openapi schema creatd from ZodDate objects to have a format type


AGalabov:
@shaikegross thank you for contributing. While the PR looks perfectly fine, we've intentionally not added this within our code base.

Please see this comment: https://github.com/asteasolutions/zod-to-openapi/pull/51#discussion_r996428248

TLDR: There is a `date-time` format but there is also just `date`. So from a generation standpoint we cannot assume on behalf of the user.

As for:
> The only way to do this without having to hack the generator is to c;hange the openapi schema creatd from ZodDate objects to have a format type
That should not be true - you can use the `.openapi` method to do:

```ts
const schema = z.date().openapi({ format: 'date-time' });
```

Please let me know if that answers your questions and works for you

PR #207, opened 2024-01-31
add autocompletion for media type

Sacramentix:
I've improve the type of ZodContentObject to provide autocompletion with most common media type without restricting the ability to put any string you want.


AGalabov:
@Sacramentix sorry this took so long. Your build is failing - can you please check it out

PR #210, opened 2024-02-26
Add support for discriminator on native enums properties

shaikegross:
Currently discriminating on a field, which changes between different NativeEnum types causes model definition to be incorrect as mapping from discriminator field doesn't map to each of the enum values.

I used the same solution that existed for ZodEnum, and applied it for ZodNativeEnum


src/openapi-generator.ts
```diff
@@ -727,6 +727,13 @@ export class OpenAPIGenerator {
         return;
       }
 
+      if (isZodType(value, 'ZodNativeEnum')) {
+        Object.values(value._def.values).forEach((enumValue: string) => {
```
AGalabov, line 731: Why do we need Object.values here as opposed to the direct usage of `value._def.values` above?

AGalabov (commented):
@shaikegross thank you for this PR - left a small question in the code. Additionally:

1. In this repo we try our best to keep everything new tested - so would you be open to adding a test for that scenario - there is a `discriminated-union.spec.ts` file where we have a z.enum test that you can use as a base.

2. Your build is failing - you have a typescript error.

PR #217, opened 2024-03-21
Refactor into classes

AGalabov:


src/transformers/index.ts
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    if (isZodType(zodSchema, 'ZodString')) {
+      return new StringTransformer().transform(zodSchema, schema =>
+        this.versionSpecifics.mapNullableType(schema, isNullable)
+      );
+    }
+
+    if (isZodType(zodSchema, 'ZodNumber')) {
+      return new NumberTransformer().transform(
```
georgyangelov, line 80: Can we extract these "new" instances as private fields, so that we don't need to create the object for every transform? Not sure how much performance difference there is but seems better to be on the safe side since this is a hot path

src/transformers/big-int.ts
```diff
@@ -0,0 +1,16 @@
...
+import { GetNumberChecks, MapNullableType } from '../types';
+
+export class BigIntTransformer {
+  transform(
+    zodSchema: ZodBigInt,
+    mapNullableType: MapNullableType,
+    getNumberChecks: GetNumberChecks
+  ) {
+    return {
+      ...mapNullableType('integer'),
```
georgyangelov, line 11: I know this probably has been discussed and has tests, but is it correct that a BigInt should have an integer representation? BigInt is arbitrary-size, so `9238745923847592837459823745982734958273498572398457298347598237459872349857239847592837459827349587239845723984759823745987239458723984572983745982734958723498572398457` is a valid value, and it can't be represented as an integer type (even 64-bit) - should be a string IMO.

georgyangelov (approved):
Looks really good! Nice one. Left one actual comment and one question

src/transformers/big-int.ts
AGalabov: Hmm you are correct. I don't remember discussing this, but it does have tests :smile:. I think I am with you on this one and I would rather implement it. However I think I'd do this in a separate PR - just for future reference and ease of searching. I still think I'd release everything in a single release.

PR #220, opened 2024-03-29
Add multi package support

danielkarso-pon:
Via: https://www.sensedeep.com/blog/posts/2021/how-to-create-single-source-npm-module.html

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@asteasolutions/zod-to-openapi",
-  "version": "7.0.0",
+  "version": "6.4.0",
   "description": "Builds OpenAPI schemas from Zod schemas",
   "main": "dist/cjs/index.js",
   "module": "dist/mjs/index.js",
--- a/package.json
+++ b/package.json
@@ -2,14 +2,21 @@
   "name": "@asteasolutions/zod-to-openapi",
   "version": "6.4.0",
   "description": "Builds OpenAPI schemas from Zod schemas",
-  "main": "dist/cjs/index.js",
   "module": "dist/mjs/index.js",
   "exports": {
     ".": {
-      "import": "./dist/mjs/index.js",
-      "require": "./dist/cjs/index.js"
+      "import": {
+        "types": "./dist/mjs/index.d.ts",
+        "default": "./dist/mjs/index.js"
+      },
+      "require": {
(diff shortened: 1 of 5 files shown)


danielkarso-pon:
PR force updated after it failed previously because of npm lint issues

georgyangelov:
Thank you for this! Let me know after you have a chance to test the package as a dependency of some project that uses ESM.

@AGalabov could you take a look as well, also do you have somewhere we can test the CJS importing before a release?

georgyangelov (approved):

package.json
```diff
@@ -1,8 +1,15 @@
 {
   "name": "@asteasolutions/zod-to-openapi",
-  "version": "6.4.0",
+  "version": "7.0.0",
```
AGalabov, line 3: There is no need to modify those manually. We have an automated release process and we'll update them accordingly

package.json
```diff
@@ -1,8 +1,15 @@
...
-  "main": "dist/index.js",
+  "main": "dist/cjs/index.js",
+  "module": "dist/mjs/index.js",
+  "exports": {
+    ".": {
+      "import": "./dist/mjs/index.js",
+      "require": "./dist/cjs/index.js"
+    }
+  },
   "types": "dist/index.d.ts",
```
AGalabov, line 13: Shouldn't the types be moved as well?

package.json
danielkarso-pon: Fixed

package.json
danielkarso-pon: Yes, that looks like something new (to me) . I followed this guide as to support nodejs and typescript for older and newer versions; https://devblogs.microsoft.com/typescript/announcing-typescript-4-7/#package-json-exports-imports-and-self-referencing

AGalabov:
@danielkarso-pon did you manage to test this? I just tried setting up some test projects. From what I see the current setup would not fully work. One thing that I saw as a problem was that the compiled js files for ESM modules are lacking the file extension `.js` at the end => the internal imports that we use in out app are "wrong" and cannot be resolved.

I think we should ultimetely try and test against the repo that was mentioned [here](https://github.com/honojs/middleware/issues/436) - that lead from hono to us

AGalabov:
@danielkarso-pon thank you once again for the contribution, ideas and the review. I will go ahead with the other PR as we agreed

PR #222, opened 2024-04-02
export ESM modules with rollup

AGalabov:
Alternative approach to ESM modules using Rollup as dev dependency

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -14,7 +14,7 @@ jobs:
     strategy:
       matrix:
         # See supported Node.js release schedule at https://nodejs.org/en/about/releases/
-        node-version: [16.x, 18.x]
+        node-version: [16.x, 18.x, 20.x]
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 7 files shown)


.github/workflows/check.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     strategy:
       matrix:
         # See supported Node.js release schedule at https://nodejs.org/en/about/releases/
-        node-version: [14.x, 16.x, 18.x]
```
AGalabov, line 17: Rollup uses `??=` internally which is not valid in node 14. I don't like the fact that we had to do this change but at the same time I do not think that it is too much of a problem :thinking: .

danielkarso-pon:
This PR looks good. I would prefer this over my PR!

.github/workflows/check.yml
georgyangelov: That's weird, are you sure it's not configurable?

AGalabov: Well that's rollup's [source code](https://github.com/rollup/rollup/blob/master/src/utils/crypto.ts#L21) :thinking: I don't think we can configure that (at least I did not find any options and don't think there are). If we wanted to we probably can introduce Babel and try and convert it to old syntax ... not sure if there's any point.

PR #229, opened 2024-05-08
anyOf contains multiple nullable values with z.enum

check24-tkilian:
The current union / enum spec produces a lot of unnecessary duplicated nullable entries. I provided a failing testcase that reproduces the issue.


AGalabov:
The test here is actually passing in 9.0.0 (fixed a while back but never closed this PR)

PR #250, opened 2024-08-08
docs: update to include information about strict usage on an object

Jafferwaffer:
Since https://github.com/asteasolutions/zod-to-openapi/pull/71 it has been possible to support `strict` mode on Zod objects, however, I couldn't find this easily in the docs.

Adding this to make it clear that by default `additionalProperties` will be set to `true`, as this may catch others out.

Happy to reword, move docs about as required.


AGalabov (approved):

AGalabov:
Thank you @Jafferwaffer

PR #279, opened 2024-12-05
Example on how to add the openapi-extension to your code

IlmariKu:
I think everyone has to figure these things on their own right now, so I wanted to add an example on how you can add the extension-method.


AGalabov (approved):
@IlmariKu thank you for the contribution

PR #282, opened 2024-12-09
regex, time and duration string types

yotamishak:
added regex, time and duration string types 

https://json-schema.org/understanding-json-schema/reference/string.html#built-in-formats

time and duration using zod built in functionality

because there is no isRegex zod function, i used
```js
this.getZodStringCheck(zodSchema, 'regex');
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -560,6 +560,8 @@ The list of all supported types as of now is:
     - `.uuid()`
     - `.email()`
     - `.url()`
+    - `.duration()`
+    - `.time()`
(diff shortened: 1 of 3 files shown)


AGalabov:
@yotamishak thank you for this. This PR became quite outdated and I never looked into it. I just saw what you proposed - the two new ones I agree with and just added them into v9.1.0 but the regex logic was incorrect. I opted for a manual implementation since the PR is very old, there are conflicts and needs an update

PR #299, opened 2025-04-24
Fix: correct regex for bigint OpenAPI pattern

habibrehmansg:
This pull request includes updates to ensure proper escaping of backslashes in regular expressions for bigint patterns. The changes affect both the test cases and the transformer logic.

**Changes to regular expression escaping:**

* [`spec/types/bigint.spec.ts`](diffhunk://#diff-9e3ba1064c96eb81b1845bd0068ad612677ca1a4726ce5bfee3c54e7fca1ea3cL7-R7): Updated the test case for the `SimpleBigInt` schema to use a properly escaped regular expression (`^\d+) for the `pattern` property.
* [`src/transformers/big-int.ts`](diffhunk://#diff-3a5c83327418db29a499287dc5bd0734330662ee69dff9af1b05ebc6b38ad68cL7-R7): Modified the `BigIntTransformer` class to use the escaped regular expression (`^\d+) in the `pattern` property when transforming a nullable type.


habibrehmansg:
Additionally, a bigint type with no default() appears as a random value in schema example value. I am unable to find out how this is generated. Ideally it should be "0".

`"id": "9429915432135503096"`

habibrehmansg:
Any chance on this getting merged?

AGalabov (approved):

AGalabov:
@habibrehmansg thank you for this contribution and sorry that I've never responded for such a long time. I will release this as a 9.1.0 release shortly.

As for your question of the default schema - I believe this comes from zod's internal default

Issue #298, opened 2025-04-15
zod v4 support


wenerme:
running with zod v4 will failed with

```
@asteasolutions/zod-to-openapi/dist/index.cjs:48
        const result = zodModifier.apply(this, args);
                                   ^

TypeError: Cannot read properties of undefined (reading 'apply')
```

v4 now support meta & jsonschema, this may make zod-to-openapi easier

colinhacks:
linking my Zod 4 PR so people can find this tracking issue: https://github.com/colinhacks/zod/pull/4074

yusukebe:
Hi @AGalabov !.

We are also working on some things to support Zod v4. We can refer to the "For library authors" section of the Zod documents:

https://v4.zod.dev/library-authors

The v4 will be released today or tomorrow, right? @colinhacks 

I know you are busy, but if the `zod-to-openapi` supports v4, we can work on about [Zod OpenAPI](https://github.com/honojs/middleware/tree/main/packages/zod-openapi). BUT, you don't hurry!

hoopyfroody:
I am right now implementing zod 3 and zod-to-openapi in our companies codebase and looking to upgrade to Zod 4 once an openapi conversion solution is ready.

If you guys need any help from a slightly above-average typescript coder, let me know :)

valerii15298:
@yusukebe zod supports `toJSONSchema` natively now, which means for openapi 3.1 which it’s compatible with JSON schema, no external dependencies needed and just zod itself is enough.

AGalabov:
@valerii15298 you are correct on that. There are lots of cases that would be covered by zod itself when it comes to basic schemas. However things like:
1. Schemas in parameters
2. Automatically linking extended schemas
3. Anything to do with actual endpoints - i.e path, params, query, body registration
cannot be dune with pure zod (and correctly so). 

As for everyone here. I am trying to squeeze in extra time to work on this. This is [the PR](https://github.com/asteasolutions/zod-to-openapi/pull/301) that I'd be working on. As of right now I've got to 162 out of 213 working tests. It is a relatively quick process but there are some things to figure out based on the new structure of zod's internals.

 I'll keep you posted

AGalabov:
# Beta version is here!

# All tests pass

<img width="570" alt="Image" src="https://github.com/user-attachments/assets/a721a38d-06c3-4f18-a19d-e9f7eea78590" />

[...]

Issue #236, opened 2024-05-29
Conditional use of "oneOf" instead of "anyOf" for particular z.union


M1r1k:
First, thank you for the great library!

I went through all related issues and PRs about but did not find anything :)

I have a case where `z.union()` is used not for different object structures but for object vs string:

```
const t = z.union([
 z.string(),
 z.object({ prop1: z.string() }),
 ]);
```

That generates OpenAPI like this:
```
anyOf:
  - type: string
  - type: object
    properties:
      prop1:
        type: string
    required:
      - prop1
```

Which ends up as 
```
export type T = Partial<string> & Partial<{ prop1: string }>;
```

And I understand that in some cases it makes sense, but when different data formats are compared I need `oneOf` instead.
So is there any "smart" way to alter OpenAPI generation? So far I see only post-processing `OpenAPIObject` as a way to conditionally alter this behaviour

Thanks in advance!

arjunyel:
@M1r1k also hitting this, what did you mean by post-processing the OpenAPIObject?

PR #301, opened 2025-05-20
Zod v4

AGalabov:


package.json
```diff
@@ -48,8 +48,8 @@
     "ts-jest": "^29.0.3",
     "typescript": "^5.2.2",
     "yaml": "^2.2.2",
-    "zod": "^3.22.0"
+    "zod": "^4.0.0-beta.20250505T195954"
```
versecafe, line 51: maybe using `^3.25.3` is safer since it allows zod 3 requirements from other things to resolve smoothly but keeps all of zod 4 accessible as

```ts
import { z } from "zod/v4";
```

grmkris:
let's get this merged 🫡 thanks

AGalabov:
@grmkris there are still some minor things to tweak to get the code to a production ready state. I hope that we should merge this, prepare an up to date README and release an official version by the end of the week

grmkris:
Any was i can help? 
i have this codebase thst id love to upgrade to v4. 
Was thinking to clone this and build locally and replace my libs with these for time being but not sure if i want to spend time on this or rather just wait

WilliamABradley:
@AGalabov here is a minimal repro: https://github.com/WilliamABradley/demo-zod4-openapi

```
Error: ID Who already exists in the registry
    at $ZodRegistry.add (file:///Users/williamabradley/Documents/GitHub/demo-zod4-openapi/node_modules/zod/dist/esm/v4/core/registries.js:13:23)
    at inst.meta (file:///Users/williamabradley/Documents/GitHub/demo-zod4-openapi/node_modules/zod/dist/esm/v4/classic/schemas.js:72:29)
    at zod.ZodType.openapi (/Users/williamabradley/Documents/GitHub/zod-to-openapi/dist/index.cjs:109:29)
    at /Users/williamabradley/Documents/GitHub/zod-to-openapi/dist/index.cjs:1074:54
    at Array.map (<anonymous>)
    at OpenAPIGenerator.generateInlineParameters (/Users/williamabradley/Documents/GitHub/zod-to-openapi/dist/index.cjs:1049:58)
    at enhanceMissingParametersError.location (/Users/williamabradley/Documents/GitHub/zod-to-openapi/dist/index.cjs:1206:83)
    at enhanceMissingParametersError (/Users/williamabradley/Documents/GitHub/zod-to-openapi/dist/index.cjs:512:16)
    at OpenAPIGenerator.getParameters (/Users/williamabradley/Documents/GitHub/zod-to-openapi/dist/index.cjs:1206:32)
    at enhanceMissingParametersError.route (/Users/williamabradley/Documents/GitHub/zod-to-openapi/dist/index.cjs:1246:69)
```

src/metadata.ts
```diff
@@ -1,55 +1,94 @@
...
+
+    const resultMetadata = {
+      ...currentMetadata,
+      ...(description ? { description } : {}),
+      ...metadata,
+    };
+
+    const totalMetadata = {
+      ...(Object.keys(_internal).length > 0 ? { _internal } : {}),
+      ...(Object.keys(resultMetadata).length > 0 ? resultMetadata : {}),
```
georgyangelov, line 29: Can this be just `...resultMetadata`?

src/transformers/intersection.ts
```diff
@@ -28,13 +28,17 @@ export class IntersectionTransformer {
...
     if (!isZodType(schema, 'ZodIntersection')) {
       return [schema];
     }
 
-    const leftSubTypes = this.flattenIntersectionTypes(schema._def.left);
-    const rightSubTypes = this.flattenIntersectionTypes(schema._def.right);
+    const leftSubTypes = isAnyZodType(schema._zod.def.left)
+      ? this.flattenIntersectionTypes(schema._zod.def.left)
+      : [];
+    const rightSubTypes = isAnyZodType(schema._zod.def.right)
```
georgyangelov, line 39: When can these not be zod types?

src/transformers/string.ts
```diff
@@ -1,11 +1,38 @@
...
+  return (
+    check._zod.def.check === 'string_format' &&
+    (check as $ZodCheckStringFormat)._zod.def.format === 'regex'
+  );
+}
 
 export class StringTransformer {
   transform(zodSchema: ZodString, mapNullableType: MapNullableType) {
-    const regexCheck = this.getZodStringCheck(zodSchema, 'regex');
+    // const regexCheck = this.getZodStringCheck(zodSchema, 'regex');
```
georgyangelov, line 25: ✂️ ?

src/transformers/string.ts
```diff
@@ -1,11 +1,38 @@
...
-    const length = this.getZodStringCheck(zodSchema, 'length')?.value;
+    const regexCheck = zodSchema.def.checks?.find(isZodCheckRegex);
+    // toString generates an additional / at the beginning and end of the pattern
+    const pattern = regexCheck?._zod.def.pattern
+      .toString()
+      .replace(/^\/|\/$/g, '');
+
+    const check = zodSchema.def.checks?.find(isZodCheckLengthEquals);
+    const length = check?._zod.def.length;
+    // const length = this.getZodStringCheck(zodSchema, 'length')?.value;`
```
georgyangelov, line 35: ✂️ ?

src/transformers/tuple.ts
```diff
@@ -10,9 +11,11 @@ export class TupleTransformer {
     mapNullableType: MapNullableType,
     mapItem: MapSubSchema
   ): SchemaObject {
-    const { items } = zodSchema._def;
+    const items = zodSchema._zod.def.items;
 
-    const schemas = items.map(mapItem);
+    const schemas = items.map(item =>
+      isAnyZodType(item) ? mapItem(item) : {}
```
georgyangelov, line 17: Is this just due to too broad TS types or is it actually possible for this to not be a zod type?

src/zod-extensions.ts
```diff
@@ -89,95 +80,93 @@ export function extendZodWithOpenApi(zod: typeof z) {
...
-    result._def.openapi = undefined;
+  // const zodOmit = zod.ZodObject.prototype.omit as any;
+  // zod.ZodObject.prototype.omit = function (this: any, ...args: any[]) {
+  //   const result = zodOmit.apply(this, args);
+  //   result.def.openapi = undefined;
 
-    return result;
-  };
+  //   return result;
+  // };
```
georgyangelov, line 171: ✂️ ?

georgyangelov (approved):
Looks good to me, thank you for all the changes!

I'm wondering if I should try to experiment with using the native Zod JSON Schema generator & utilizing their new registry classes. We can provide that in some next version as an alternative to our own generator - not sure the Zod one will be able to support everything we do here. In particular, I'm not sure how it handles the unwrapping of metadata we do currently (preserving the metadata across calls to .optional() and such).

In any case, even if that works well, we should continue to support our own generator at least for some time, so I'd vote for releasing it as is (modulo any bugs we encounter and fix beforehand).

Note: this should be a major version bump, possibly a .beta for a while?

georgyangelov:
@AGalabov on @WilliamABradley 's problem, looks like calling `schema.meta(val)` (internally does `z.globalRegistry.add(schema, val)` on an object that's already in the registry is an error. That's not obvious from the docs and I thought we could override as well, but from the error it seems like that's the way it works.

I noticed that we have `z.globalRegistry.has(` so we can check for it and I'm thinking we can do a `remove` + `add` in that case.

georgyangelov:
Or, maybe, we can mutate the metadata object in the registry? Not sure if that's allowed...

georgyangelov:
Thinking about this more, I'm not sure we should be storing stuff in the `.meta()` or the global zod registry in this version at all. This may potentially pollute the context for other libraries that rely on what is stored there.
For example, generating a JSON schema with zod's own z.jsonSchema will put ALL fields from the registry into the result (including our `_internal` stuff). There's a lot of potential for conflicts.

I believe we should create our own registry and store our internal data + whatever is passed to `openapi(_, { here })` in it. That way we keep separation and don't globally affect what someone may have put in the global registry.

WilliamABradley:
@georgyangelov this library could still read from the global registry though right?

I'd like to put the schema `id`, `title`, `pattern`, `examples`, etc. using the `.meta()` call, and then have it flow through into the OpenAPI schema.

Are the `.def` or `._zod` properties on the schema class the locations for library authors to store their own internal metadata?

georgyangelov:
@WilliamABradley 

> this library could still read from the global zod registry though right?
Yes, but I'm not sure to what extent we should. `description` is obvious (and we're already reading it), but `id` - not really currently. The way it works right now is we need to register the zod schema in our own registry (OpenAPIRegistry) for the docs generation to work. We can probably safely support `title`, `pattern` and `examples` as well.

But currently you won't be able to do `.meta({ id: 'something' }` instead of `.openapi('something')`.

We'll be able to do that if we rely on Zod's own JSON Schema generator, but we aren't doing that here because: (1) it's going to be more work both to implement and make sure we bring the cases we've been handling that possibly Zod doesn't handle and (2) it's going to be a breaking change in one way or another - we don't want to pull the rug on everyone. Also Zod doesn't seem to support OpenAPI 3.0 (they support OpenAPI 3.1.1 only - the latest version which includes the latest JSON schema), so some people won't be able to use it.

> Aren't the .def or ._zod properties on the schema class the locations for library authors to store their own internal metadata?

I don't think they are, the `def` provides description of the shape of objects and `_zod` (having an underscore) is "private".
That is, unless I'm missing something 😄  Please point me to a reference if so

src/metadata.ts
```diff
@@ -1,55 +1,94 @@
...
+
+    const description = currentMetadata?.description ?? schema.description;
+
+    const resultMetadata = {
+      ...currentMetadata,
+      ...(description ? { description } : {}),
+      ...metadata,
+    };
+
+    const totalMetadata = {
```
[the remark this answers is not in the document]
versecafe: lmfao this is an incredible PR review

src/metadata.ts
```diff
@@ -1,55 +1,94 @@
...
+    metadata?: ZodOpenApiFullMetadata
+  ): ZodOpenApiFullMetadata | undefined {
+    const currentMetadata = schema.meta() as ZodOpenApiFullMetadata | undefined;
+
+    const _internal = {
+      ...currentMetadata?._internal,
+      ...metadata?._internal,
+    };
+
+    const description = currentMetadata?.description ?? schema.description;
```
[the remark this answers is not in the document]
AGalabov: I think I've tried that but it might've been before my fixed. I'll double check

src/metadata.ts
AGalabov: Yes it cat. It used to be `resultMetadata.metadata` and I've missed the opportunity

spec/modifiers/meta.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+describe('meta', () => {
+  it('should preserve metadata through the meta modifier', () => {
+    const schema = z
+      .string()
+      .meta({
+        description: 'Some description',
+        deprecated: true,
+        example: 'test',
+      })
+      .openapi('SomeString');
```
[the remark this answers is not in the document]
versecafe: can meta do subfields ie

```ts
const schema = z.object({ something: z.string() }).meta({ openapi: { id: "SomeString" } });
```

since the nice to have of the `.openapi()` was it was easy to notice where things came from in the specgen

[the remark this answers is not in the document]
kernwig: @versecafe It'd be allowed, but Typescript wouldn't help you any.

```ts
export interface JSONSchemaMeta {
    id?: string | undefined;
    title?: string | undefined;
    description?: string | undefined;
    examples?: $output[] | undefined;
    [k: string]: unknown;
}
export interface GlobalMeta extends JSONSchemaMeta {
}
```

The `id` property already there serves this same purpose for JSON Schema. I generate both JSON Schema and OpenAPI spec, so in the interest of DRY I want to use the same for both.

We could use `satisfies` to add type checking though:

```ts
interface ZodOpenAPIMetadata<T = unknown> extends z.core.JSONSchemaMeta {
  openapi?: {
    default?: T;
  };
}

export const userIdSchema = z
  .string()
  .min(1)
  .meta({
    id: "UserId",
    description: "User Identity subject identifier (a user's unique ID).",
    openapi: { 
      format: "bad", // TS2353: Object literal may only specify known properties, and format does not exist in type
      default: "0" 
    },
  } satisfies ZodOpenAPIMetadata<string>);
```

Although it saves us from having to call `extendZodWithOpenApi(z)` on every entry point, I'm not thrilled by it. It could be present as an _option_ though.  As stated before, **I** rarely use `.openapi` for anything other than setting `id`, so using the existing metadata `id` handles 98% of my uses.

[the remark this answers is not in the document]
versecafe: didn't realize id was actually specified and not just a feature of the [k: string]: unknown, makes sense to have it just use `.meta({ id })`

AGalabov:
@WilliamABradley I applied the changes suggested by @georgyangelov and used a custom registry for our logic. The repro you gave us is now passing - that is there are no errors. I've not checked the result in detail. 

Please feel free to test out the latest version (commit) and let us know if everything seems alright. @georgyangelov I think we are ready for a second review pass now.

P.S. I've not tried working on reading the `id` from meta.

src/metadata.ts
```diff
@@ -4,7 +4,8 @@ import { ZodOpenAPIMetadata, ZodOpenApiFullMetadata } from './zod-extensions';
 import { isNil, omit, omitBy } from './lib/lodash';
 import { ParameterObject, ReferenceObject, SchemaObject } from './types';
 
-const zodToOpenAPIRegistry = z.registry<ZodOpenApiFullMetadata<any>>();
+const zodToOpenAPIRegistry = z.registry<ZodOpenApiFullMetadata>();
```
WilliamABradley, line 7: Could we export this for introspection?

WilliamABradley:
@AGalabov as you say, the repro is now working. It is not hoisting the components, but I'm assuming that is related to reading the `id` from meta?

Interestingly, it is adding the `id` in the inline spec:

```yaml
info:
  title: demo-zod4-openapi
  version: 1.0.0
openapi: 3.1.0
components:
  schemas: {}
  parameters: {}
paths:
  /my-route:
    get:
      operationId: MyRoute
      deprecated: false
      requestBody:
        required: true
        content:
          application/json:
            schema:
              type: object
              properties:
                who:
                  type: string
                  id: Who
              required:
                - who
              id: MySpecialParam
      responses:
        "200":
          description: Success
          content:
            application/json:
              schema:
                type: object
                properties:
                  who:
                    type: string
                    id: Who
                  message:
                    type: string
                required:
                  - who
                  - message
                id: MySpecialResponse
  /my-route/:who:
    get:
      operationId: MyRouteWithPathParams
      deprecated: false
      parameters:
        - schema:
            type: string
            id: Who
          required: true
          name: who
          in: path
      responses:
        "200":
          description: Success
          content:
            application/json:
              schema:
                type: object
                properties:
                  who:
                    type: string
                    id: Who
                  message:
                    type: string
                required:
                  - who
                  - message
                id: MySpecialResponse
webhooks: {}
```

AGalabov:
@WilliamABradley @kernwig I agree on the fact that id should be mapped to our internal refId. I've just pushed a commit that does that. Please check it out.

However on the title -> summary mapping I am not sure that I agree. First of all a title is valid in OpenAPI v3.1 (something that we do support). And then a summary is not really a SchemaObject's property but a path/response one.

kernwig:
> However on the title -> summary mapping I am not sure that I agree. First of all a title is valid in OpenAPI v3.1 (something that we do support).

Ah, I see where summary vs title is used in OpenAPI. My confusion... `title` should carry through as `description` does.

WilliamABradley:
Cool, I'll check out the new changes

hoopyfroody:
Has my comment here: https://github.com/asteasolutions/zod-to-openapi/issues/298#issuecomment-2912515734 been seen and/or fixed? :)

AGalabov:
@WilliamABradley since you've been so active here (and one of the few that use the local setup to a forked repo). Can you confirm that the id registration works as expected for you. I am planning to release a new beta version very soon (with updated README.md) and then close the main issue for zod v4 support #301. Then leave it for a bit to see how things go and eventually make it the official release.

flipvh:
Thanks for all of this @AGalabov , amazing work! I will try to see if I can help testing using our [hono/zod-openapi](https://github.com/honojs/middleware/issues/1177) but I guess that requires first some work on this hono middleware itself.

kernwig:
I'm now using 8.0.0-beta.2, and it's working quite nicely for me. I was even able to remove the `extendZodWithOpenApi(z)` call, since I don't have any need to call `z.openapi()` anymore.

WilliamABradley:
@AGalabov it is working perfectly now!

```ts
info:
  title: demo-zod4-openapi
  version: 1.0.0
openapi: 3.1.0
components:
  schemas:
    TestId:
      type: prefixid
      examples:
        - test_12345
    Who:
      type: string
      pattern: ^[a-zA-Z]+$
      examples:
        - World
    MySpecialResponse:
      type: object
      properties:
        _id:
          $ref: "#/components/schemas/TestId"
        who:
          $ref: "#/components/schemas/Who"
        message:
          type: string
      required:
        - who
        - message
    MySpecialParam:
      type: object
      properties:
        _id:
          $ref: "#/components/schemas/TestId"
        who:
          $ref: "#/components/schemas/Who"
      required:
        - who
  parameters: {}
paths:
  /my-route:
    get:
      operationId: MyRoute
      deprecated: false
      requestBody:
        required: true
        content:
          application/json:
            schema:
              $ref: "#/components/schemas/MySpecialParam"
      responses:
        "200":
          description: Success
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/MySpecialResponse"
  /my-route/{who}:
    get:
      operationId: MyRouteWithPathParams
      deprecated: false
      parameters:
        - schema:
            type: prefixid
            examples:
              - test_12345
          required: false
          name: _id
          in: path
        - schema:
            $ref: "#/components/schemas/Who"
          required: true
          name: who
          in: path
      responses:
        "200":
          description: Success
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/MySpecialResponse"
webhooks: {}
```

colinhacks:
Please open issues against Zod if you find any unexpected/buggy behavior in `z.toJSONSchema()`

AGalabov:
@WilliamABradley that is correct. We are not using the toJSONSchema with this iteration and I am not sure if we would be able to fully integrate it any time soon. As for oneOf/anyOf this is something that has been required for a long time. See here:
* https://github.com/asteasolutions/zod-to-openapi/issues/236 and
* https://github.com/asteasolutions/zod-to-openapi/discussions/227

You could see an answer for it here: https://github.com/asteasolutions/zod-to-openapi/pull/163#issuecomment-1649450148. At this point in time we are not going to support a decision between the two - other than using the `.openapi()` method to manually input stuff.

As for the issue with the order of invocations - I will try and take a look soon

AGalabov:
@WilliamABradley actually the problem is pretty much the same one:
1. Our generator generates the schema as "anyOf"
2. Then we compare whether there is any additional "properties" (OpenAPI wise) to be added to the resulting schema.
3. In your case the registered APIIssuePathPart has `anyOf: []`, but the generated items get generated with `anyOf: [{ type: 'number' }, { type: 'string' }]` so this "new" information gets added to the `allOf`

In the case where you do not register it - it simply gets inlined and the output is the same as the one of the initial schema (anyOf: []).

We cannot change this as this is "valid" behaviour. The only thing that can be fixed is the anyOf/oneOf that I've already shared thoughts on.

georgyangelov (approved):
🎉 🎉 🎉

grmkris:
let's go?

PR #308, opened 2025-06-13
add support for conditional oneOf instead of anyOf for unions

AGalabov:
A potential fix of #236  and  #227.

It adds on top of the intended implementaion in #163 as it adds both a general configuration and a one of(f) /no pun intended/ scenarios for particular unions through the usage of `openapiAs` (only for unions).

If we agree on the implementation I can add some description for it in the README.md.


src/zod-extensions.ts
```diff
@@ -181,4 +188,17 @@ export function extendZodWithOpenApi(zod: typeof z) {
 
     return result;
   };
+
+  zod.ZodUnion.prototype.openapiAs = function (
```
georgyangelov, line 192: Hmm, not 100% sure on including another function. Can this be done with the `.openapi` itself? Maybe as an additional parameter so that it doesn't clash with openapi fields?

georgyangelov (approved):

Issue #290, opened 2025-01-13
z.never() in request body schema throws exception


IbrahimMohammed47:
When I register a path, where body schema is `z.never()` or contains a nested attribute that is `z.never`, the following exception is thrown:
```
node:internal/process/promises:289
            triggerUncaughtException(err, true /* fromPromise */);
            ^

[UnhandledPromiseRejection: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "#<UnknownZodTypeError>".] {
  code: 'ERR_UNHANDLED_REJECTION'
}
```

Here is an example that causes the exception:
```js
registry.registerPath({                           // registry: OpenAPIRegistry
    method: 'put',
    path: '/todos/{id}',
    request: {
      params: verifyIdSchema,
      body: {
        content: {
          'application/json': {
            schema: z.object({ x: z.never() })    // same if z.never() is deeply nested in the schema object
          }
        }
      }
    },
    responses: {}
  })
```

Same if I use `z.undefined()` as well

Is this a bug or am I missing something?

AGalabov:
@IbrahimMohammed47 the error you are mentioning is not the correct one I don't think. When I tried this locally with a minimal example I see that the error is:

```
UnknownZodTypeError {
  message: 'Unknown zod object type, please specify `type` and other OpenAPI props using `ZodSchema.openapi`.',
  data: { currentSchema: { typeName: 'ZodNever' }, schemaName: undefined }
}
```

and that is expected. There isn't really anything that we can translate this to in terms of "open api specification". There is no equivalent to this because undefined for example is exactly the same as not having that key (in terms of HTTP communication). And never is just a typescript type - it does not represent any particular form of data.

If you still want to use this as your validation from zod. You can just create the openapi definition manually by passing a `type`. I am not sure what that would be for those two cases but you can do for example:

```ts
schema: z.object({ x: z.never().openapi({ type: 'null'}) })
```

Hope that helps. I will be closing the ticket with that comment.

PR #344, opened 2025-11-15
feat: add `void` & `never` support

nickthegroot:
Currently, `void` and `never` are failing with the following error:

```
UnknownZodTypeError {
  message: 'Unknown zod object type, please specify `type` and other OpenAPI props using `schema.openapi`.',
  data: {
    currentSchema: { type: 'never', error: [Function: error] },
    schemaName: undefined
  }
}
```

This was brought up in #290, although the closure felt a little premature: https://github.com/asteasolutions/zod-to-openapi/issues/290#issuecomment-2646129379.

> There isn't really anything that we can translate this to in terms of "open api specification"

I'm completely in support of this idea, although would advocate for the library gracefully handling these instead of erroring out. This is what this PR attempts to do. After all, if OpenAPI is meant to express what _should_ be in the request, what _shouldn't_ be in there has no relevance!

**AI Disclosure**: I used copilot to generate an initial plan and framework, particularly for the tests. All code was subsequently personally reviewed and edited.

### DIFF
--- a/spec/types/never.spec.ts
+++ b/spec/types/never.spec.ts
@@ -1,0 +1,144 @@
+import { z } from 'zod';
+import { expectSchema } from '../lib/helpers';
+
+describe('never', () => {
+  describe('in object properties', () => {
+    it('skips never properties in objects', () => {
+      const schema = z
+        .object({
+          id: z.string(),
+          impossible: z.never(),
+          name: z.string(),
+        })
+        .openapi('TestObject');
+
+      expectSchema([schema], {
+        TestObject: {
+          type: 'object',
+          properties: {
+            id: {
+              type: 'string',
+            },
+            name: {
+              type: 'string',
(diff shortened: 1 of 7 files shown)


AGalabov:
@nickthegroot thank you for the contribution. However I still stand besides my statement https://github.com/asteasolutions/zod-to-openapi/issues/290#issuecomment-2646129379. 

In particular your PR showcases it. Whilst using `{}` might sound like a good idea to represent never/void this is actually incorrect mapping. In OpenAPI the lack of type (and in particular `{}`) means "anything" which is exactly the opposite of what never/void would imply if used for validation.

I do believe that the error is the correct way to handle it - i.e it is unknown behaviour that the user should decide how to handle. The error is descriptive enough and pinpoints the expected resolution.

Because of that I would have to close this PR.

Issue #327, opened 2025-07-25
Support for `.prefault()` in zod 4?


florianmartens:
I'm running zod 4 and `zod-to-openapi` v8.0.0.

I'm getting this error for a value with a new zod `prefault`.

```
{
  message: Unknown zod object type, please specify `type` and other OpenAPI props using `schema.openapi`.
  currentSchema: {
    type: 'prefault'
    ...
  }
}
```

Any plans to support `prefault`? What would be the best way around this error?

0xb4lamx:
I've tried to address this in https://github.com/asteasolutions/zod-to-openapi/pull/349. Would appreciate a look!

PR #349, opened 2025-12-05
feat: support zodv4 prefault new type

0xb4lamx:
**Context:**

Zod 4 [introduced `.prefault()`](https://zod.dev/api?id=prefaults) to replicate Zod 3's `.default()` behavior, where the default value is passed *into* the parser (e.g. for transforms), whereas the new `.default()` short-circuits and returns the value directly.

Since OpenAPI has no concept of "pre-parse" vs "short-circuit" defaults (it only has a single `default` field for missing values), this PR maps both Zod concepts to the OpenAPI `default` field.

**This PR includes:**

support for Zod 4's new `prefault` type, and ensures that the default value is correctly extracts regardless of which method is used, prioritizing `.default()` to match runtime precedence:

1.  **`src/lib/zod-is-type.ts`**:
    *   Added `ZodPrefault` to `ZodTypes` and `ZodTypeKeys` definitions.
    *   Ensured `ZodPrefault` is recognized as a distinct Zod type by the internal type checking logic.
2.  **`src/metadata.ts`**:
    *   Updated schema traversal logic (`collectMetadata`, `unwrapUntil`) to handle `ZodPrefault` wrappers, preventing metadata loss when `prefault()` is used.
    *   Updated `getDefaultValue` to support `prefault` values as a fallback for the OpenAPI `default` field.
        *   **Precedence:** `ZodDefault` is checked first. In Zod 4, `.default()` short-circuits the parsing process and returns the value directly if the input is `undefined`.
        *   **Fallback:** If no `ZodDefault` is found, `ZodPrefault` is used. `prefault` ("pre-parse default") passes the default value *into* the schema parsing logic _(replicating Zod 3's default behavior)_.

related issue https://github.com/asteasolutions/zod-to-openapi/issues/327

### DIFF
--- a/README.md
+++ b/README.md
@@ -630,6 +630,7 @@ The list of all supported types as of now is:
 - `ZodBoolean`
 - `ZodDate`
 - `ZodDefault`
+- `ZodPrefault`
 - `ZodDiscriminatedUnion`
   - including `discriminator` mapping when all Zod objects in the union are registered with `.register()` or contain a `refId`.
 - `ZodEffects`
--- a/spec/modifiers/prefault.spec.ts
+++ b/spec/modifiers/prefault.spec.ts
@@ -1,0 +1,44 @@
+import { z } from 'zod';
+import { expectSchema } from '../lib/helpers';
+
+describe('prefault', () => {
+  it('supports prefault values', () => {
+    const schema = z
+      .string()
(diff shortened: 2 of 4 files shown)


AGalabov (approved):
@0xb4lamx thank you for this contribution! I'll release this shortly

PR #357, opened 2026-01-08
Enum components are being generated as nullable when the property holding it is nullable

### LINKED ISSUE #258: Enum components are being generated as nullable when the property holding it is nullable
### Description
I have the following scenario:

```typescript
const Brand = z.enum(['Ford', 'Toyota', 'Tesla']).openapi('Brand');

const Car = z
    .object({
        brand: Brand.nullable(),
        model: z.string(),
    })
    .openapi('Car');
```
 
### Expected result
I expected the property `Car.brand` to be a nullable `Brand`

```yaml
components:
  schemas:
    Brand:
      type: string
      enum:
        - Ford
        - Toyota
        - Tesla
    Car:
      type: object
      properties:
        brand:
          oneOf:
            - type: 'null'
            - $ref: '#/components/schemas/Brand'
        model:
          type: string
```

### Actual result
The spec generated declares `Brand` as nullable, instead of making the property `Car.brand` nullable

```yaml
components:
  schemas:
    Brand:
      type:
        - string
        - 'null'
      enum:
        - Ford
        - Toyota
        - Tesla
    Car:
      type: object
      properties:
        brand:
          $ref: '#/components/schemas/Brand'
        model:
          type: string
```

### More info

I understand the rationale behind the way this was generated. The problem is that `Brand` could be reusable, and in some places it might not be nullable. Is there a workaround this?

giskard-prime-5906:
Fixes #258

### DIFF
--- a/spec/types/enum.spec.ts
+++ b/spec/types/enum.spec.ts
@@ -31,4 +31,71 @@ describe('enum', () => {
       },
     });
   });
+
+  it('supports nullable enum references in objects', () => {
+    const Brand = z.enum(['Ford', 'Toyota', 'Tesla']).openapi('Brand');
+
+    const Car = z
+      .object({
+        brand: Brand.nullable(),
+        model: z.string(),
+      })
+      .openapi('Car');
+
+    expectSchema([Brand, Car], {
+      Brand: {
+        type: 'string',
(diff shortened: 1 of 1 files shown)


AGalabov:
@giskard-prime-5906 this is tests only. I see another open PR for this issue so I will turn my attention to it.

PR #374, opened 2026-03-31
backport #373 to zod v3 version

HannesGitH:
This cannot be merged onto master, a specific legacy zod-v3 branch (or similar) must be created from v7.3.4 and then this PR can be applied to that branch

I would be happy if you'd `npm publish --tag legacy` this, although you said you wont, so I dont need to use my fork. 👉👈


AGalabov:
@HannesGitH thank you for this. However I really don't want to go down the rabbit hole of backporting stuff onto a legacy branch. It would make maintenance a bit harder as every now and then I might have to review code agains a legacy version from months/years  back.  So I would have to decline this, sorry

PR #388, opened 2026-06-25
feat: add OpenApiGeneratorV32 for OpenAPI 3.2

### LINKED ISSUE #385: Proposal: OpenAPI 3.2 support — OpenApiGeneratorV32 subclass vs. a version option?
👋 Thanks for this library — it's the engine behind `@hono/zod-openapi` and a good chunk of our API tooling.

OpenAPI 3.2 is finalized and I'd like to add support for it. It touches the generator surface, so I wanted to agree on the shape before opening a PR.

## Why this matters

3.2 is almost entirely additive at the document-structure level. The feature we actually need is `itemSchema` on a media type, which finally lets you type the items of a streaming response (`text/event-stream`, `application/jsonl`, `application/json-seq`) instead of falling back to `z.string()`. A few other additive fields come along for free: optional response `description` + `summary`, the `query` method, and `itemEncoding` / `prefixEncoding`.

## Current gap

3.2 did **not** change the JSON Schema dialect — it's still 2020-12, same as 3.1. So `OpenApiGeneratorV31`'s schema conversion already emits 3.2-valid schemas; there's no dialect work. The gap is just: a generator that stamps `3.2` and accepts the new fields, plus converting `itemSchema` (when it's a Zod schema) so it `$ref`s like `schema` already does. The only external piece is `openapi3-ts` shipping the 3.2 types (there's an open PR for that).

## The design question

Two ways to expose it, and I'd rather match your preference before writing it up:

```ts
// (a) a new generator class, mirroring V3 / V31
new OpenApiGeneratorV32(registry.definitions).generateDocument({ openapi: '3.2.0', ... })

// (b) a version flag on the existing generator that permits the 3.2 fields
new OpenApiGeneratorV31(registry.definitions).generateDocument({ openapi: '3.2.0', ... })
```

(a) matches the existing `OpenApiGeneratorV3` / `OpenApiGeneratorV31` split and keeps each version's types separate; (b) is fewer moving parts but blurs the version boundary. I lean (a) for consistency since the dialect lets `V32` reuse the `V31` specifics, but it's your call.

Either way it stays opt-in — `V3` and `V31` output is unchanged.

## Desired behavior

```ts
const Event = registry.register('Event', z.object({ message: z.string() }))

registry.registerPath({
  method: 'get',
  path: '/events',
  responses: {
    200: {
      description: 'Event stream',
      content: { 'text/event-stream': { itemSchema: Event } },
    },
  },
})
// content['text/event-stream'].itemSchema === { $ref: '#/components/schemas/Event' }
```

Happy to open the PR — I have it working locally against a 3.2-typed `openapi3-ts`. Just let me know if you'd prefer (a) or (b).

bookernath:
Resolves #385.

Adds `OpenApiGeneratorV32` for generating OpenAPI `3.2.x` documents, following the subclass approach (option **a**) that @AGalabov agreed with on the issue.

## Approach

OpenAPI 3.2 keeps the 3.1 JSON Schema dialect (2020-12), so schema generation is identical to 3.1. Rather than duplicate that logic, the new generator reuses `OpenApiGeneratorV31Specifics` and only layers on the document-structure fields introduced in 3.2:

- **`itemSchema`** on a media type — describes each item of a sequential stream (`text/event-stream`, `application/jsonl`, `application/json-seq`, …). A Zod schema is converted and registered exactly like `schema` (so it `$ref`s a component); a raw `SchemaObject`/`ReferenceObject` is passed through.
- **`itemEncoding` / `prefixEncoding`** on media types
- **Reusable media types** via `components.mediaTypes`, with media type `$ref`s passed through from `content`
- **`summary`** and an optional **`description`** on responses
- the **`query`** HTTP method on routes

## Dependency

Requires the `oas32` module from `openapi3-ts`, added in metadevpro/openapi3-ts#158 and published in `4.6.0`. The dependency range is bumped `^4.1.2` → `^4.6.0` accordingly, and the lockfile is synced so `npm ci` stays green in CI.

## Notes

- The lazy transformer needed a cast because `openapi3-ts >= 4.2` added an optional `$ref` to `SchemaObject`, so `'in'` no longer narrows the union.
- Webhook generation runs before final component assembly so webhook-only named schemas and `itemSchema`s resolve correctly.

## Tests

- New `spec/v3.2.spec.ts` covering the 3.2-specific behavior, reusable media type refs, webhook-only item schemas, and encoding pass-through
- Full suite green locally: build, tests, type tests (`tsd`), and prettier

### DIFF
--- a/spec/v3.2/media-types.spec.ts
+++ b/spec/v3.2/media-types.spec.ts
@@ -1,39 +1,8 @@
 import { z } from 'zod';
-import { OpenAPIRegistry } from '../src/openapi-registry';
-import {
-  OpenApiGeneratorV32,
-  OpenAPIObjectConfigV32,
-} from '../src/v3.2/openapi-generator';
+import { OpenAPIRegistry } from '../../src/openapi-registry';
+import { generateV32Document } from './helpers';
 
-const testDocConfig: OpenAPIObjectConfigV32 = {
-  openapi: '3.2.0',
-  info: { version: '1.0.0', title: 'Test API' },
-};
-
-function generate(registry: OpenAPIRegistry) {
-  return new OpenApiGeneratorV32(registry.definitions).generateDocument(
-    testDocConfig
-  );
-}
-
-describe('OpenApiGeneratorV32', () => {
-  it('keeps the 3.2 openapi version supplied by the config', () => {
-    const document = generate(new OpenAPIRegistry());
-
-    expect(document.openapi).toEqual('3.2.0');
-  });
-
-  it('emits the same JSON Schema dialect as 3.1 (type arrays for nullable)', () => {
-    const registry = new OpenAPIRegistry();
-    registry.register('NullableString', z.string().nullable());
-
-    const document = generate(registry);
-
-    expect(document.components?.schemas?.['NullableString']).toEqual({
-      type: ['string', 'null'],
-    });
-  });
-
+describe('OpenAPI 3.2 media types', () => {
   describe('itemSchema', () => {
     it('converts a Zod itemSchema and registers it as a component ref', () => {
       const registry = new OpenAPIRegistry();
@@ -58,7 +27,7 @@ describe('OpenApiGeneratorV32', () => {
         },
       });
 
-      const document = generate(registry);
+      const document = generateV32Document(registry);
       const content =
         document.paths?.['/events']?.get?.responses?.['200']?.content?.[
           'text/event-stream'
@@ -92,7 +61,7 @@ describe('OpenApiGeneratorV32', () => {
         },
       });
 
-      const document = generate(registry);
+      const document = generateV32Document(registry);
       const content =
         document.paths?.['/lines']?.get?.responses?.['200']?.content?.[
           'application/jsonl'
@@ -125,7 +94,7 @@ describe('OpenApiGeneratorV32', () => {
         },
       });
 
-      const document = generate(registry);
+      const document = generateV32Document(registry);
       const content =
         document.paths?.['/raw']?.get?.responses?.['200']?.content?.[
           'application/json-seq'
@@ -153,7 +122,7 @@ describe('OpenApiGeneratorV32', () => {
         },
       });
 
-      const document = generate(registry);
+      const document = generateV32Document(registry);
       const content =
         document.paths?.['/both']?.get?.responses?.['200']?.content?.[
           'multipart/mixed'
@@ -191,7 +160,7 @@ describe('OpenApiGeneratorV32', () => {
         },
       });
 
-      const document = generate(registry);
+      const document = generateV32Document(registry);
       const content =
         document.webhooks?.['eventCreated']?.post?.responses?.['200']
           ?.content?.['application/jsonl'];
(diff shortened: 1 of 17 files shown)


src/v3.2/openapi-generator.ts
```diff
@@ -0,0 +1,78 @@
...
+
+    return {
+      ...config,
+      ...baseDocument,
+      webhooks: this.webhookRefs,
+    } as OpenAPIObject;
+  }
+
+  generateComponents(): Pick<OpenAPIObject, 'components'> {
+    return this.generator.generateComponents() as Pick<
```
AGalabov: I don't think this cast should be needed?

spec/v3.2/media-types.spec.ts
AGalabov, line 1: I have quite a few comments on this file:
1. Thank you for having some extensive tests, however:
2. This test does not follow our "convention" of using the `expectSchema` function
3. It does not follow our "grouped" tests. I think some of those should go into already pre-existing files. And some of those should be separated into smaller files. If needed - I would agree to group them as a v3.2 folder or something like that

src/openapi-generator.ts
```diff
@@ -99,7 +99,7 @@ export class OpenAPIGenerator {
   private paramRefs: Record<string, ParameterObject> = {};
   private pathRefs: Record<string, PathItemObject> = {};
   private rawComponents: {
-    componentType: keyof ComponentsObject;
+    componentType: string;
```
AGalabov: Why is this needed? If anything - shouldn't it be `ComponentTypeKey`?

src/openapi-generator.ts
```diff
@@ -133,7 +133,8 @@ export class OpenAPIGenerator {
   }
 
   private buildComponents(): ComponentsObject {
-    const rawComponents: ComponentsObject = {};
+    const rawComponents = {} as ComponentsObject &
```
AGalabov: I believe if you just update the type in `types.ts` this cast should not be needed?

src/openapi-generator.ts
```diff
@@ -745,11 +747,14 @@ export class OpenAPIGenerator {
...
 
+    // `description` is required on 3.0/3.1 ResponseObjects but optional on the
+    // shared ResponseConfig (it is optional since OAS 3.2). Keeping it correct
+    // per target version is the user's responsibility, so we cast here.
     if (!headers) {
       return {
         ...rest,
         ...responseContent,
-      };
+      } as ResponseObject;
```
AGalabov: I agree that with the changes in 3.2 we should allow for optional descriptions (considering we have a single entry point). However casting is not the correct way as if anyone actually uses this on their end programatically it would be incorrect. We should change the `types.ts` as well here - and make sure that our ResponseObject supports optional descriptions

src/openapi-registry.ts
```diff
@@ -130,14 +192,27 @@ export type OpenAPIComponentObject =
...
+type ComponentValue<T, K extends PropertyKey> = T extends unknown
+  ? K extends keyof T
+    ? NonNullable<T[K]> extends Record<string, infer V>
+      ? V
+      : never
+    : never
+  : never;
+
+export type ComponentTypeKey = Exclude<KeysOfUnion<ComponentsObject>, number>;
+export type ComponentTypeOf<K extends ComponentTypeKey> = ComponentValue<
```
AGalabov: I couldn't really trace what the idea was with those type constructors. Can you please explain

src/openapi-registry.ts
```diff
@@ -237,7 +312,7 @@ export class OpenAPIRegistry {
       type: 'component',
       componentType: type,
       name,
-      component,
+      component: component as OpenAPIComponentObject,
```
AGalabov: I don't like that we have a cast here either

src/openapi-registry.ts
bookernath: Replaced with `ComponentTypeMap` which I think is more self explanatory?

src/v3.0/openapi-generator.ts
```diff
@@ -30,10 +30,13 @@ export class OpenApiGeneratorV3 {
     return {
       ...config,
       ...baseData,
-    };
+    } as OpenAPIObject;
```
AGalabov: Once again I don't think this is needed, is it? (same goes for the v3.1 file)

src/v3.0/openapi-generator.ts
```diff
@@ -30,10 +30,13 @@ export class OpenApiGeneratorV3 {
...
   }
 
   generateComponents(): Pick<OpenAPIObject, 'components'> {
-    return this.generator.generateComponents();
+    return this.generator.generateComponents() as Pick<
```
AGalabov: And this?

src/openapi-generator.ts
```diff
@@ -133,33 +141,36 @@ export class OpenAPIGenerator {
   }
 
   private buildComponents(): ComponentsObject {
```
AGalabov, line 143: Maybe something like this would work as well. It narrows down the type casting strictly to the raw components that are user input and can somewhat be ignored from the type check.

```ts
private buildComponents(): ComponentsObject {
    type RawComponent = Record<string, OpenAPIComponentObject>;

    const rawComponents: Record<string, RawComponent> = {};
    this.rawComponents.forEach(({ componentType, name, component }) => {
      rawComponents[componentType] ??= {};
      rawComponents[componentType][name] = component;
    });

    // This is entirely based on the user's control so it is okay to cast it.
    const rawSchemas = (rawComponents['schemas'] ?? {}) as Record<
      string,
      SchemaObject | ReferenceObject
    >;

    const allSchemas = {
      ...rawSchemas,
      ...this.filteredSchemaRefs,
    };

    const schemas =
      this.options?.sortComponents === 'alphabetically'
        ? sortObjectByKeys(allSchemas)
        : allSchemas;

    // This is entirely based on the user's control so it is okay to cast it.
    const rawParameters = (rawComponents['parameters'] ?? {}) as Record<
      string,
      ParameterObject | ReferenceObject
    >;

    const allParameters = {
      ...rawParameters,
      ...this.paramRefs,
    };

    const parameters =
      this.options?.sortComponents === 'alphabetically'
        ? sortObjectByKeys(allParameters)
        : allParameters;

    return { ...rawComponents, schemas, parameters };
  }
```

AGalabov (commented):
There are some more casts that I do not like. Otherwise it is getting closer to what I envision

bookernath:
Agreed that the casts in the v3.0 and v3.1 generators weren’t necessary, removed those. I also narrowed the remaining component casts to only the raw user-provided schema and parameter buckets while keeping generated values type-checked.

AGalabov:
@bookernath thank you for this contribution! 🚀

PR #390, opened 2026-07-30
Read metadata through zod's core registry so zod/mini schemas work

josher8a:
## Problem

`getMetadataFromRegistry` calls `zodSchema.meta()`, which `zod/mini` schemas do not have:

```
TypeError: zodSchema.meta is not a function
    at Metadata.getMetadataFromRegistry (src/metadata.ts:209)
    at Metadata.getRefId
    at OpenAPIGenerator.generateSchemaWithRef
```

## Change

```diff
-    const general = zodSchema.meta();
+    const general = z.core.globalRegistry.get(zodSchema);
```

Same store, so classic is unchanged. `z` was already imported.

`spec/zod-mini.spec.ts` added. 319 passing (was 316), lint clean.

## Not included

Read path only. Still classic-only: `schemaWithRefId` (openapi-registry.ts:311) calls `.openapi(refId)`, and `getMetadataFromRegistry` early-returns when `internal` is absent, so a zod-native `id` never becomes `refId`. That second one changes existing output, so your call.


josher8a:
**On the two follow-ups in the description:**

**Runtime:** extracting the variant-agnostic half of `.openapi()` (merge, clone via `z.core.clone`, record) into `Metadata.applyMetadata` lets `schemaWithRefId` and `registerParameter` skip the prototype. Mini then emits `{"$ref":"#/components/schemas/Wallet"}` instead of throwing. 320 passing. Can send separately.

**Types:** the remaining blocker.

```
error TS2345: 'ZodMiniObject<...>' is not assignable to 'ZodType<...>'
```

Retyping `register`, `applyMetadata` and the `Metadata` methods to `$ZodType`: 3 errors. Widening `OpenAPIDefinitions.schema`: 49 of 53 suites, since it fans out to every generator and transformer. Wide but mechanical, roughly 84 `ZodType` and 23 `ZodObject` across 21 files, incremental with the suite gating each step.

Want it? Fine to stop at the one-liner, or add the runtime change, or do the migration. Or `id` to `refId` first if you would rather settle that behaviour question before the mechanical work.

